PR #229, opened 2014-07-25
Muller potential example dataset

rmcgibbo:


rmcgibbo:
This is outdated. I used some of the code to implement #450, but the layout is a little different.

Issue #437, opened 2015-01-07
Slow `_transition_counts` performance


synapticarbors:
In testing MSMBuilder3 with one of my systems, I noticed that the `_transition_counts` method was extremely slow for me. I have ~1000 trajectory sequences, each with between 1000-5000 frames. Granted I'm on an older workstation, but calling `_transition_counts` on my sequence of assignments took ~57 seconds. Digging around in the code, I noted that you were doing something which I had done previously in one of our WESTPA tools -- using scipy sparse matrices to accumulate counts. It's actually very expensive to build the `coo_matrix` and convert it to a dense matrix over and over again for each sequence. If you first accumulate all of the `to_states` and `from_states` into a list of arrays and then concatenate them at the end, calling the `coo_matrix` constructor once, you can get drastically better performance.

The following modification to `_transition_counts` runs in about 1.4 seconds on my machine on the same data set, giving the same results:

```python
def _transition_counts_new(sequences, lag_time=1, sliding_window=True):

    if (not sliding_window) and lag_time > 1:
        return  _transition_counts([X[::lag_time] for X in sequences], lag_time=1)

    classes = np.unique(np.concatenate(sequences))
    contains_nan = (classes.dtype.kind == 'f') and np.any(np.isnan(classes))
    contains_none = any(c is None for c in classes)

    if contains_nan:
        classes = classes[~np.isnan(classes)]
    if contains_none:
        classes = [c for c in classes if c is not None]

    n_states = len(classes)

    mapping = dict(zip(classes, range(n_states)))
    mapping_is_identity = np.all(classes == np.arange(n_states))
    mapping_fn = np.vectorize(mapping.get, otypes=[np.int])
    none_to_nan = np.vectorize(lambda x: np.nan if x is None else x,
                               otypes=[np.float])

    counts = np.zeros((n_states, n_states), dtype=float)
    
    _transitions = []
    for y in sequences:
        y = np.asarray(y)
        from_states = y[: -lag_time: 1]
        to_states = y[lag_time::1]

        if contains_none:
            from_states = none_to_nan(from_states)
            to_states = none_to_nan(to_states)

        if contains_nan or contains_none:
            # mask out nan in either from_states or to_states
            mask = ~(np.isnan(from_states) + np.isnan(to_states))
            from_states = from_states[mask]
            to_states = to_states[mask]

        if (not mapping_is_identity) and len(from_states) > 0 and len(to_states) > 0:
            from_states = mapping_fn(from_states)
[...]

PR #438, opened 2015-01-08
Modified _transition_counts method in msm to increase performance

synapticarbors:
This PR addresses https://github.com/msmbuilder/msmbuilder/issues/437 by modifying `_transition_counts` to only create `coo_matrix` at the end of the method rather than once for every sequence. 

The following plots show the timings and speed-ups of the revised method vs the original version in master for a variety of trajectory lengths, numbers of trajectories and number of states. The corresponding gist can be found at:

http://nbviewer.ipython.org/gist/synapticarbors/6b7fc9ba004bbab42f32

Over the range of values considered here, the revised method does well compared to the current implementation. I maybe should have tried a set of assignments that resulted in a higher sparsity though. 

The changes pass all of the tests for the module and I also checked that I get the same results as the original implementation. I'm happy to revise this PR further if needed.

timings


rmcgibbo:
LGTM. I'll wait for the tests to run, and then merge.

rmcgibbo:
Grrr. Test failure is related to https://github.com/omnia-md/conda-recipes/issues/80.

rmcgibbo:
Tested locally.

Issue #436, opened 2015-01-07
Potentially confusing scaling of transition count matrix


synapticarbors:
In digging through the source code, I noticed something that could be potentially confusing to users. In defining the graph in `strongly_connected_subgraph` there is a comparison done between the counts and the `weight` argument to the method:

https://github.com/msmbuilder/msmbuilder/blob/master/msmbuilder/msm/core.py#L203

This is clearly documented in the docstring. However `_transition_counts()` doesn't actually return the observed transitions, but rather the observed transitions scaled by the lag time:

https://github.com/msmbuilder/msmbuilder/blob/master/msmbuilder/msm/core.py#L330

I'm not sure what the rationale is for this, but what it means is that you can have fractional counts that fall below the default weight cut-off of `1` and thus not contribute an edge in the graph when determining the largest ergodic set. So for example:

```
T, mapping = _transition_counts([[0,1,0,1,0,1,0,1]], lag_time=3)
```

returns:

```
[[ 0.          1.        ]
 [ 0.66666667  0.        ]]
```

rather than what I would expect from the count matrix:

```
[[ 0.  3.]
 [ 2.  0.]]
```

Passing the fractional count matrix to `_strongly_connected_subgraph` with the default `weight=1` results in 2 strongly connected components rather than 1 if you actually had `T * lag_time` as the count matrix.

Maybe there is a good modeling reason for this (?), but it's not clear that this would happen from the doc strings and it seems unexpected in general.

kyleabeauchamp:
I believe the scaling was introduced to correct the error estimates on rate and transition matrices when using a sliding window (e.g. non-independent counts).

mpharrigan:
I think you're right, but this is confusing re: cutoff

rmcgibbo:
> I believe the scaling was introduced to correct the error estimates on rate and transition matrices when using a sliding window (e.g. non-independent counts).

Yes, this hasn't been described in the literature yet, but it's something I'm working on. I think I can demonstrate quantitatively that the scaling gives substantially improved error estimators, at least for simple systems where we can generate `N` independent collections of trajectories for the same system. Note that scaling doesn't change the MLE.

> Passing the fractional count matrix to _strongly_connected_subgraph with the default weight=1 results in 2 strongly connected components rather than 1 if you actually had T * lag_time as the count matrix.

[...]

Issue #433, opened 2014-12-17
Implement APM clustering


rmcgibbo:
- New [paper](http://pubs.acs.org/doi/pdf/10.1021/ct5007168) on ASAP from the Huang group at HKUST.
- Looks fairly computationally cheap.
- Potentially nice for situations where the diffusion constant is highly position dependent (e.g. multi-body systems with an bound vs. unbound state)
- Their code is GPL, which isn't compatible w/ our LGPL. I sent an email to Xuhui asking if they would consider (dual/re)-licensing under a compatible license.

mpharrigan:
closed by #748

PR #439, opened 2015-01-08
use observed rather than scaled transition counts for ergodic trim

synapticarbors:
This is a simple fix to https://github.com/msmbuilder/msmbuilder/issues/436 which simply multiplies the count matrices by the lag time when calculating the strong connected component for the ergodic trim. It's maybe not the most elegant solution, but it at least temporarily makes the algorithm consistent with the docstrings and removes the unexpected behavior that I encountered. Additionally I'm now getting comparable results for a system that I had previously analyzed with MSMBuilder2 using similar settings.

### DIFF
--- a/msmbuilder/msm/bayesmsm.py
+++ b/msmbuilder/msm/bayesmsm.py
@@ -180,7 +180,7 @@ class BayesianMarkovStateModel(BaseEstimator, _MappingTransformMixin):
 
         if self.ergodic_cutoff >= 1:
             self.countsmat_, mapping2 = _strongly_connected_subgraph(
-                raw_counts, self.ergodic_cutoff, self.verbose)
+                self.lag_time * raw_counts, self.ergodic_cutoff, self.verbose)
             self.mapping_ = _dict_compose(mapping, mapping2)
         else:
             self.countsmat_ = raw_counts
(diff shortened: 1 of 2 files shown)


rmcgibbo:
It's not clear to me that this is an improvement. Assume that we're using sliding window, and say, `lag time=10`, but the shortest timescale in the system is >>10. Then, every entry in the counts matrix is going to be a multiple of 10 -- every transition is observed redundantly `lag_time` times. It seems to me that the ergodic cutoff should increase proportionally as well. (Or equivalently, the cutoff should be applied to `counts / lag_time`).

synapticarbors:
In the current scheme where the count matrix is actually the number of counts divide by the lag time you need to observe at least the lag time worth of counts to not have your edge pruned. I agree with your rationale, but at least for the system I'm looking at, the difference results in either having a relatively well connected state graph or not. Part of this may be that I'm looking at a binding process where my state space has regions with very different diffusive timescales (maybe https://github.com/msmbuilder/msmbuilder/issues/433 would help here), and certain bin-to-bin transitions don't have many counts in one direction due to a non-equilibrium starting distribution and less than ideal sampling.

Since the code currently doesn't allow you to trim with an  `ergodic_cutoff` < 1, this seemed like a reasonable way of preserving the observed connectivity or at least introducing the flexibility to go down to keeping edges for which at least one real transition is observed.

rmcgibbo:
Okay. I'll merge this for now. Let's keep thinking about the best way to deal with this issue going forward.

PR #449, opened 2015-01-17
Asymptotic eigenvalue and timescale uncertainty estimators

rmcgibbo:
The whole appendix of this paper [1] is supposed to show an algorithm for calculating the partial derivatives of the eigenvalues of the transition matrix w.r.t. the transition matrix elements, `\frac{\partial \lambda_k}{\partial T_{ij}`, but I don't think authors didn't really do it the right way. I'm sure the way its written in the paper is correct, but they skipped the _much_ simpler solution, which is just `\frac{\partial \lambda_k}{\partial T_{ij} = U_{ik} V_{jk}`, where U and V are the matrices with the left and right eigenvectors respectively as columns. I guess this is only true of the the transition matrix satisfies detailed balance, but I think that's the situation of interest.

[1] [Calculation of the distribution of eigenvalues and eigenvectors in Markovian state models for molecular dynamics](http://scitation.aip.org/content/aip/journal/jcp/126/24/10.1063/1.2740261)

### DIFF
--- a/msmbuilder/msm/msm.py
+++ b/msmbuilder/msm/msm.py
@@ -657,8 +657,9 @@ Timescales:
             for i in range(self.n_states_):
                 ui = self.countsmat_[:, i]
                 wi = np.sum(ui)
-                covariance = wi*np.diag(ui) - np.outer(ui, ui)
-                sigma2[k] += dLambda_dT[i].dot(covariance).dot(dLambda_dT[i]) / (wi**2*(wi+1))
+                cov = wi*np.diag(ui) - np.outer(ui, ui)
+                quad_form = dLambda_dT[i].dot(cov).dot(dLambda_dT[i])
+                sigma2[k] += quad_form / (wi**2*(wi+1))
         return np.sqrt(sigma2)
 
     def uncertainty_timescales(self):
--- a/msmbuilder/msm/msm.py
+++ b/msmbuilder/msm/msm.py
@@ -643,6 +643,8 @@ Timescales:
            the distribution of eigenvalues and eigenvectors in Markovian state
            models for molecular dynamics." J. Chem. Phys. 126.24 (2007): 244101.
         """
+        if self.reversible_type is None:
+            raise NotImplementedError('reversible_type must be "mle" or "transpose"')
 
         n_timescales = min(self.n_timescales, self.n_states_ - 1)
         if n_timescales is None:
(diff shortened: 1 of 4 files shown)


rmcgibbo:
This plot is an attempt to reproduce-in-spirit Figure 2 of the Hinrichs paper, comparing the distribution of a bunch of eigenvalues and timescales generated by randomly sampling from a dirichlet with the analytic approximation for the distributions.

test_msm_uncertainty_plots

kyleabeauchamp:
Cool, will try to take a look.

msmbuilder/msm/msm.py
```diff
@@ -627,3 +627,60 @@ def draw_samples(self, sequences, n_samples, random_state=None):
...
+            n_timescales = self.n_states_ - 1
+        u, lv, rv = self._get_eigensystem()
+
+        sigma2 = np.zeros(n_timescales + 1)
+        for k in range(n_timescales + 1):
+            dLambda_dT = np.outer(lv[:, k], rv[:, k])
+            for i in range(self.n_states_):
+                ui = self.countsmat_[:, i]
+                wi = np.sum(ui)
+                cov = wi*np.diag(ui) - np.outer(ui, ui)
```
rmcgibbo: I think this loop over `i` could be vectorized with `einsum`.

rmcgibbo: Unsuccessful.

jhprinz:
I implemented this at some point for EMMA to get the partial derivatives for eigenvalues. If I am not mistaken then this is independent of any detailed balance constrained and the derivation is quite simple.
I checked the code by comparison with a simple numeric differentiation. So this is correct. It uses exclusively numpy, but there might still be better implementations.

You can find the whole set of sensitivities in 

https://github.com/markovmodel/PyEMMA/blob/devel/pyemma/msm/analysis/dense/sensitivity.py#L133



```
def eigenvalue_sensitivity(self, k):
    """ 
    calculate the sensitivity matrix for eigenvalue k given transition matrix T.
    
    Parameters
    ----------
    T : np.ndarray shape = (n, n)
        Transition matrix
    k : int
        eigenvalue index for eigenvalues order descending
        
    Returns
    -------
    x : ndarray, shape=(n, n)
        Sensitivity matrix for entry index around transition matrix T. Reversibility is not assumed.
        
    """
        
    eValues, rightEigenvectors = np.linalg.eig(self.T)
    leftEigenvectors = np.linalg.inv(rightEigenvectors)    
    
    perm = np.argsort(eValues)[::-1]

    rightEigenvectors=rightEigenvectors[:,perm]
    leftEigenvectors=leftEigenvectors[perm]
    
    sensitivity = np.outer(leftEigenvectors[k], rightEigenvectors[:,k])
    
    return sensitivity
```

It contains also sensitivities for forward and backward committor, eigenvectors, implied timescales, etc... Especially the backward committor is kind of messy.

rmcgibbo:
Cool.

PR #450, opened 2015-01-19
Muller potential dataset

rmcgibbo:
Updated #229 to more modern project layout.


kyleabeauchamp:
Looks reasonable, seems like github is too stupid to prettyprint the `muller.py` file though

PR #452, opened 2015-01-21
Add ergodic trimming in ContinuousTimeMSM

rmcgibbo:
I was getting some very odd results with GMRQ, where the test scores seemed sort of unstable. For example, I'd have a training set score of ~ -0.1, and over randomized shuffle splits, my test scores would be ~ -0.2, -0.2, -0.3, -100. I tracked down one of the erroneous shuffle-splits, and made it a test: `test_score_2`. The issue is related to zeros in the count matrix. Adding ergodic trimming (matching the behavior of `MSM`) seems to fix it up.


rmcgibbo:
This still doesn't fix the instability issue.

PR #453, opened 2015-01-21
[ENH] Switch to transmat style GMRQ

rmcgibbo:


kyleabeauchamp:
works for me

rmcgibbo:
I do not really understand the different between these formulations, but under cross validation unless things are very well behaved (really big training and test sets that are very 'close'), this formulation is much more stable than the rate matrix version. My intuition is that there's an argument you could make from perturbation theory to explain the effect.

Issue #458, opened 2015-01-27
The dataset object does not implement ```keys``` correctly


schwancr:
When doing ```dataset.__get__(i)```, i corresponds to the i'th item **after** sorting.

But when iterating through the dataset the keys are never sorted, which means anything that does: ```for item in dataset``` (unless the keys come out sorted {certainly not the case for HDF5 datasets}) the trajectories get scrambled. 

https://github.com/msmbuilder/msmbuilder/blob/master/msmbuilder/dataset.py#L308-L312

schwancr:
Thanks to @sryckbos for finding this problem.

schwancr:
@rmcgibbo @mpharrigan maybe an arduino can fix this issue?

rmcgibbo:
Fixed by #460.

PR #459, opened 2015-01-27
possible fix for #458 there are probably other ways to do this too

schwancr:
We can either require that the implementation of the Datasets sorts the keys correctly, or we can sort it in the _BaseDataset class. 

Does anyone have a preference? Are there cases, where the sorted command in the _BaseDataset will not work?

### DIFF
--- a/msmbuilder/dataset.py
+++ b/msmbuilder/dataset.py
@@ -173,7 +173,7 @@ class _BaseDataset(Sequence):
         return self.set(i, x)
 
     def __iter__(self):
-        for key in self.keys():
+        for key in sorted(self.keys(), key=_keynat):
             yield self.get(key)
 
     def keys(self):


rmcgibbo:
Sorting in the base sounds fine. Can you add a test?

rmcgibbo:
No, nevermind. You should not sort in `__iter__`. It's just marching through `keys()`. Instead, you should sort the output of `keys`. This is already done in `NumpyDirDataset`. It needs to be also done in `HDF5Dataset`.

rmcgibbo:
Actually, I'm not really sure.

schwancr:
Depends what we want ```.keys()``` to do. If it's emulating a dict, then a python user will know that these are not supposed to be sorted numerically.

rmcgibbo:
Making the keys orderable makes sense.

- `keys()[i]` should be the same as `.get(i)` and the `i`th thing returned by iter.
- we should put in some tests on this sort behavior.

schwancr:
I think that makes sense. So, each dataset implementation has to implement keys such that this is the case. I'll make a simple test

rmcgibbo:
Fixed by #460.

PR #460, opened 2015-01-28
Fix sort order

rmcgibbo:
This test now passes. It did not pass before.

```
def test_order_1():
    with tempdir():
        with dataset('ds1.h5', 'w', 'hdf5') as ds1:
            for i in range(20):
                ds1[i] = np.random.randn(10)
            assert list(ds1.keys()) == list(range(20))

        with dataset('ds1/', 'w', 'dir-npy') as ds1:
            for i in range(20):
                ds1[i] = np.random.randn(10)
            assert list(ds1.keys()) == list(range(20))
```

cc #458, #459.

### DIFF
--- a/msmbuilder/dataset.py
+++ b/msmbuilder/dataset.py
@@ -83,6 +83,8 @@ def dataset(path, mode='r', fmt=None, verbose=False, **kwargs):
 
 
 def _guess_format(path):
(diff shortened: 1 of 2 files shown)


rmcgibbo:
travis failure is gateway timeout.

PR #469, opened 2015-02-12
remove value error for trajectories that are too short in tICA._fit

schwancr:
@sryckbos was using tICA with osprey on some FAH data. Some trajectories are too short depending on the correlation lag time, but the code throws an error instead of failing silently.

So, tICA._fit just silently ignores the data. Alternatively, we could print a warning, but throwing an exception is the wrong behavior.

### DIFF
--- a/msmbuilder/decomposition/tica.py
+++ b/msmbuilder/decomposition/tica.py
@@ -14,6 +14,7 @@ from ..base import BaseEstimator
 from ..utils import check_iter_of_sequences
 from sklearn.base import TransformerMixin
 from sklearn.utils import array2d
+import warnings
 
 __all__ = ['tICA']
 
@@ -363,6 +364,7 @@ class tICA(BaseEstimator, TransformerMixin):
 
         # We don't need t scream and shout here. Just ignore this data.
         if not len(X) > self.lag_time:
+            warnings.warn("length of data (%d) is too short for the lag time (%d)" % (len(X), self.lag_time))
             return
 
         self.n_observations_ += X.shape[0]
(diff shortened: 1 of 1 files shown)


kyleabeauchamp:
I guess my vote is to have some form of warning / printing, as I think that could help prevent some common mistakes.

rmcgibbo:
How about a line in the summarize() printout?

schwancr:
That would mean, we'd need tICA to have an attribute that accumulates trajectories that we've skipped. Does that really make sense?

rmcgibbo:
I think so. We currently have `n_sequences_`, which is the number of sequences that contribute to the accumulated matrices. This would just be `len(X) - n_sequences_`, no?

schwancr:
What if someone calls partial_fit twice?

schwancr:
Idc, I think it's fine to accumulate it

schwancr:
one sec

rmcgibbo:
> What if someone calls partial_fit twice?

:) I think we can figure out how to handle this correctly.

schwancr:
haha yea, i just mean that we need to have an attribute that stores either the number we've skipped or the number total we've seen, can't get around that.

But you're right, we're already keeping track of n_sequences, so I'll add an n_skipped to track the number that we skip

mpharrigan:
This breaks saving the resulting dataset. The modification returns None which can't be serialized to `.h5`

mpharrigan:
I'm also concerned that if we implement this poorly, we won't be able to back out which md trajectory resulted in a particular tica trajectory (relevant for pulling structures)

schwancr:
You can still transform the too short trajectories, you just can't fit the model with them

schwancr:
> This breaks saving the resulting dataset. The modification returns None which can't be serialized to .h5

I don't think this is true. What is returning None?

mpharrigan:
Ok I guess you're right

mpharrigan:
Is this ready to merge then?

schwancr:
appveyor fails, what does that mean?

PR #470, opened 2015-02-14
[ENH] Change rate matrix parameterization

rmcgibbo:
This changes the semantics of the free parameters in the rate matrix optimization. Previously, the first `n*(n-1)/2` elements in theta parameterized the *log* of the (off-diagonal) symmetric rate matrix elements. This changes it so that no log transformation is applied.

The problem with the log parameterization is that it prevents the optimizer from converging to solutions with `K_{ij} = 0`. Previously, we were using a hack that involved guessing which elements might be zero and then reducing the size of the active set during optimization.

The potential advantage of the log transform was that it prevents the optimizer from needing to "see" the constraint that the off-diagonal rate matrix elements must be non-negative. But this is actually not really necessary. L-BFGS-B can handle these box constraints just fine.

Changing the parameterization simplifies the code a lot, which is nice, and empirically leads to sparser MLE solutions for a couple datasets.

There is a problem with the Hessian though currently. It has to do with the fact, I think, that the multivariate delta method only applies when the MLE is in the interior of the feasible set, not at a boundary. This means we're formally SOL, I think, and need to work around this by only computing the Hessian of the "active" variables. I'm working on this now.


rmcgibbo:
I think I was wrong about the problems with the Hessian. The switch in the parameterization doesn't really change the accuracy of the Hessian.

rmcgibbo:
This kind of breaks the Bayesian HMC sampler. I'm not sure that's so bad though -- it was never really production-ready to begin with.

jchodera:
Just out of curiosity: When you change parameterization, are you using a Jacobian-based prior to ensure the resulting posterior distribution from HMC is independent of parameterization?

rmcgibbo:
For the Bayesian variant, I changed the prior so that the posterior was invariant, so yes, that meant modifying a couple jacobian terms.

But, the bayesian variant is still broken, because in pyhmc I haven't implemented support for constraints on the parameter space. With this PR, many of the parameters are constrained to the real half line, but pyhmc doesn't know about that.

rmcgibbo:
There needs to be an elastic collision with the constraint I guess. In the log-space parameterization obviously this wasn't necessary, because there were no constraints

rmcgibbo:
Does anyone want to code-review this? I'm ready to merge it.

kyleabeauchamp:
I made a 15 minute pass through and saw no obvious red flags.  Obviously I'd need to read through the whole codebase and do the math to do a fully rigorous review, but that won't happen in the next week or two.

rmcgibbo:
Okay. That's good enough for me. I've looked at this quite a bit, and I think it's an improvement. The paper is coming :)

PR #482, opened 2015-02-25
Added some windows notes to installation.rst

kyleabeauchamp:
Obviously this is a bit rough but probably better than nothing.


rmcgibbo:
Do you actually need to cd into scripts?

Otherwise looks good.

kyleabeauchamp:
So on my box, the install didn't seem to set the PATHs automatically

rmcgibbo:
Even from inside the "Anaconda Command Prompt" shell? wierd.

Issue #485, opened 2015-02-25
MSMBUILDER 3.0 tutorial


akiikius:
Is there any way to get a quick run through of msmbuilder 3.0 that is similar to the msmbuilder 2.8 tutorial. Am new to using msmbuilder and am having a hard time figuring out the processes on the website. A quick 1,2,3 list would help. Thanks!

mpharrigan:
There's this:
http://msmbuilder.org/latest/commandline.html

It looks slightly out of date. I'll run through it to check

PR #487, opened 2015-02-25
Small modifications to command line tutorial

mpharrigan:
cc #485 

 - Changed the title to be more obvious that it is a tutorial
 - Small fixes to filenames to correspond to use of "datasets"

### DIFF
--- a/doc/commandline.rst
+++ b/doc/commandline.rst
@@ -1,12 +1,13 @@
 .. _commandline:
 .. highlight:: bash
 
-Command Line Interface (CLI)
-============================
+Command Line Tutorial
+=====================
 
-MSMBuilder is really designed (at this point) as a python library. It's much more
-powerful that way, and not any less convenient (especially when you're deailing
-with large datasets). But there is a limited CLI as well.
+MSMBuilder is designed as a python library and a command-line program.  The
+API can be much more powerful, and should be easy for researchers familiar
+with python to pick up. The command-line interface wraps the most common
+use cases.
 
 Tutorial
 --------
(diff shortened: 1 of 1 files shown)


rmcgibbo:
We should consider making command line tutorials in the form of IPython notebooks that use a bash backend. https://github.com/takluyver/bash_kernel

mpharrigan:
Ipython bash backend would be a long term thing. That requires ipy 3 (not released yet)

rmcgibbo:
I think it's going to be released this week. Not for immediate use, yes, but something like that might be a nice tool for tutorials that involve showing command line code, in the way that the ipython notebook is nice for tutorials that involve showing people python code.

rmcgibbo:
I got the bash kernel working with `pip install --pre ipython`, which gives 3.0.0rc1.

See e.g.

http://nbviewer.ipython.org/gist/rmcgibbo/d80b21971909799952f6

mpharrigan:
Is bash kernel included with ipy3?

rmcgibbo:
No, need to install it separately. But most users will just by copy-pasting, not actually downloading and running the .ipynb files I would guess.

kyleabeauchamp:
I'll leave it up to you guys to decide whether this will be easy to maintain given the pre-release status of the two prerequisites.

rmcgibbo:
We don't need to do this now. It's just something to watch out for going forward.

zjlhjun:
Yes, I think the tutorial is very important. Following the examples, we will learn it very fast. But, I can not find the tutorial in the version 3.1.

PR #489, opened 2015-02-27
Changed link for talk slides.

kyleabeauchamp:
Also, I have updated the slides to work with MSMB 3.1 (https://github.com/kyleabeauchamp/MSMBuilderTalk)


kyleabeauchamp:
PS: the py27 failure appears to be some sort of NAN issue with the score function applied to rate models.

rmcgibbo:
> PS: the py27 failure appears to be some sort of NAN issue with the score function applied to rate models.

Yes, I've seen that. It's fixed in #470, which is ready to merge.

PR #500, opened 2015-03-13
Normalize features when performing a union

mpharrigan:
Per feedback in group meeting, not doing this could affect tICA regularization.

I also don't think passing a list of paths to `dataset()` should cause features to be unioned by default. This is probably surprising for a user. I put in a deprecation warning and we can change it later


mpharrigan:
Hold on, I implemented this wrong.
We should compute the variance among *all* the trajectories

mpharrigan:
Changed it

mpharrigan:
Review?

rmcgibbo:
Is this best implemented within the dataset object? It seems like maybe a separate step / transformer.

mpharrigan:
Perhaps

mpharrigan:
How would this work as a transformation?
```python
ds1 = dataset("protein")
ds2 = dataset("water")
union = FeatureUnioner(normalize=True)
trajs = union.fit_transform(zip(ds1, ds2))
```

Thoughts?

kyleabeauchamp:
This seems pretty reasonable.  I suppose the following would also be an example of this API?

```
ds1 = dataset("phi")
ds2 = dataset("psi")
ds3 = dataset("chi")
union = FeatureUnioner(normalize=True)
trajs = union.fit_transform(zip(ds1, ds2, d3))
```

rmcgibbo:
I think this would be fine.

PR #502, opened 2015-03-16
Fix ratematrix loglikeilhood bug

kyleabeauchamp:
Fixes this bug:

```
pyflakes ratematrix.py 
ratematrix.py:203: undefined name 'loglikelihood'
```


kyleabeauchamp:
PS: this bug is causing the other PRs to fail travis as well.

PR #507, opened 2015-03-16
Refactor msm.sample() to core.py as a Mixin, now available in ratematrix.

kyleabeauchamp:


rmcgibbo:
Are you sure that the `sample` function for `ContinuousTimeMSM` should be the discrete-time version, not the gillespie? IMHO, for the continous time model, you should use a continuous-time sampler by default.

kyleabeauchamp:
Whatever we call it, IMHO we want `msm.sample()` and `rate.sample()` to be compatible, right?  And given that `msm.sample()` already exists, this was the least `breaking` api change.

rmcgibbo:
I'm not sure they need to be compatible.

rmcgibbo:
Or perhaps rephrase "compatible" -> "have the same return value semantics".

kyleabeauchamp:
I'm happy to change the naming convention to whatever people agree upon.

kyleabeauchamp:
OK, I also moved `draw_samples` into this MixIn.

rmcgibbo:
> I'm happy to change the naming convention to whatever people agree upon.

I don't think anyone besides us is going to have a strong opinion.

kyleabeauchamp:
OK, I guess my vote is to leave the names as-is for now and worry about it when someone gets around to implementing the continuous time gillespie

rmcgibbo:
We can always implement them one at a time, but it makes sense to pick the names first, so that we don't have to break the API when we want to add the next one.

My thought is that the continuous-time sampler is the natural sampler for a continuous-time model. Lag time is really only a feature of the parameterization here, but not the model itself. The (state_sequence, waiting_times) tuple is the natural (complete) description of a continuous time process, whereas its snapshot at a discrete time interval isn't a complete description of the process, because there can be unobserved jumps between the snapshots and you don't get the waiting times.

So IMO it makes sense to make the "natural" sampler for each model get the most "natural" method name.

kyleabeauchamp:
I suppose I would vote to eliminate any "natural" sampler idea and just go with explicit names for discrete time and continuous time sampling, rather than re-use the names.

rmcgibbo:
That's just an argument why they should be named `sample` and `sample_discrete` or something.

kyleabeauchamp:
How about sample_continuous and sample_discrete

rmcgibbo:
Sure. That's fine.

kyleabeauchamp:
Cool.  This also clears up a naming clash with `MSM.sample()` and `bayesMSM.sample()`, which are totally different things

kyleabeauchamp:
OK, renamed `sample()` as `sample_discrete()`.

rmcgibbo:
LGTM.

kyleabeauchamp:
OK, I guess we can merge after travis passes.  Things just passed on my machine:

```
Ran 321 tests in 87.029s

OK (SKIP=11)
TEST END: msmbuilder-dev-3.0.0-np19py27_0
```

kyleabeauchamp:
So Travis thinks this is good.  The appveyor failures appear to be some stupid MSVS compiler problems.

PR #511, opened 2015-03-23
Allow appending to datasets

mpharrigan:
At least for `dir-npy`

### DIFF
--- a/msmbuilder/dataset.py
+++ b/msmbuilder/dataset.py
@@ -40,8 +40,9 @@ def dataset(path, mode='r', fmt=None, verbose=False, **kwargs):
     ----------
     path : str
         The path to the dataset on the filesystem
-    mode : {'r', 'w'}
-        Open a dataset for reading or writing
+    mode : {'r', 'w', 'a'}
+        Open a dataset for reading, writing, or appending. Note that
+        some formats only support a subset of these modes.
     fmt : {'dir-npy', 'hdf5', 'mdtraj'}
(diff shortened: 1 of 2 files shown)


rmcgibbo:
test?

mpharrigan:
added a test

mpharrigan:
P.S. my use case for this is featurizing in parallel

rmcgibbo:
in the test: what about appending to one of the values associated with a
key that's already in the dataset?

On Mon, Mar 23, 2015 at 11:13 AM, Matthew Harrigan <notifications@github.com

mpharrigan:
What's the expected behavior? I'd think overwrite

kyleabeauchamp:
Unrelated question: are you using this functionality via the command line or via python?  I'm mostly just curious whether I should be trying to emulate your workflows here, but there's not so much in terms of "idiom" examples in the docs.

mpharrigan:
Right now I'm doing it via IPython.parallel. And featurizing many microseconds in < 2 minutes

kyleabeauchamp:
Thanks, good to know.

rmcgibbo:
> What's the expected behavior?

I'm not sure. I figured that we should probably decide though.

mpharrigan:
With `w` it's been overwrite

rmcgibbo:
Okay. Looks good. the docstrings mention this.

mpharrigan:
Gah py2 issues

PR #512, opened 2015-03-23
Add items() to dataset

mpharrigan:
It works like `dict.items()`

### DIFF
--- a/msmbuilder/dataset.py
+++ b/msmbuilder/dataset.py
@@ -188,6 +188,10 @@ class _BaseDataset(Sequence):
         # the same ordering convention for the indices / items.
         raise NotImplementedError('implemeneted in subclass')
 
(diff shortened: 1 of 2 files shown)


mpharrigan:
Something that I came across during this: slice notation doesn't work for non-contiguous indices. The dataset acts in all other ways like a dictionary with integer keys. What would you think about changing the slice notation to slice the keys that actually exist:

```
ds[0] = np.array()
ds[2] = np.array()
ds[4] = np.array()
arr_list = ds[:] # currently raises an exception
assert len(arr_list) == 3
assert ds[::2] == [ds[0], ds[4]]
```

Too unintuitive?

mpharrigan:
Rebased

mpharrigan:
Travis failure is timeout cc #517

PR #513, opened 2015-03-23
Allow PCA.fit() to accept generators

mpharrigan:


mpharrigan:
Test failure is an http timeout

Issue #514, opened 2015-03-23
KMedoids cluster_ids_ is in concatenated space


mpharrigan:
The indices would ideally be of the `form (traj_i, frame_i)`

rmcgibbo:
Yeah, that would be ideal. `KMedoids` could override `fit()` from
`_KMedoids` to call super, and then remap this.

mpharrigan:
This functionality might be best in `MultiSequenceClusterMixin`?

rmcgibbo:
Maybe? But not every clustering has the concept of `cluster_ids_`. e.g. for kmeans or something, no remapping is necessary.

rmcgibbo:
Maybe putting a method/staticmethod that is capable of doing the remapping into `MultiSequenceClusterMixin` makes sense, but then actually "manually" calling it from `KMedoids`  with `self.cluster_ids_` would make sense.

mpharrigan:
I agree

rmcgibbo:
:+1:

mpharrigan:
Can I de-double-underscore `MultiSequenceClusterMixin.__lengths`

rmcgibbo:
yeah sure. whatever makes sense.

PR #515, opened 2015-03-24
ENH: Split cluster_ids_ in *KMedoids

mpharrigan:
cc #514


rmcgibbo:
can you make sure in the test that this works when the input trajectories are not the same length?

mpharrigan:
Good point

mpharrigan:
Tests updated. Also seems to work on my actual dataset

rmcgibbo:
Is there any way to make the test actually exact though?

rmcgibbo:
like, check that the referenced cluster centers, in coordinate space, are equal to `X[_KMedoids().fit(X).cluster_ids_]`?

mpharrigan:
Oh yeah that would be better

mpharrigan:
Travis failure is an http timeout

PR #520, opened 2015-03-25
RegularSpatial exposes the indices of the cluster representatives

pfrstg:


mpharrigan:
We should add a call to the method introduced in #515

rmcgibbo:
LGTM.

rmcgibbo:
FWIW, my experience is that RegularSpatial is really slow. I think the main loop needs to be done entirely in compiled code.

rmcgibbo:
A similar change, for I think the same reason, was recently added to PyEMMA in https://github.com/markovmodel/PyEMMA/pull/134.

I would prefer not to maintain any more C that interacts with the python C API though, with `PyArg_ParseTuple`, `Py_XDECREF` and friends.

pfrstg:
Updated the code to use _split_indices

mpharrigan:
Actually I think the call to `_split_indices` should be in `RegularSpacial.fit()` not `_RegularSpacial.fit()`, which operate on a single or multiple trajs, rep.
```python
        MultiSequenceClusterMixin.fit(self, sequences)
        self.cluster_ids_ = self._split(self.cluster_ids_)
        return self
```

pfrstg:
Moved to RegularSpatial.fit
Thought note the documentation on _RegularSpatial is now kind of wrong for that class, but right for RegularSpatial. LMK if you want me to make separate attributes or anything like that.

Should I be concerned about the appveyor failures?

rmcgibbo:
> Thought note the documentation on _RegularSpatial is now kind of wrong for that class, but right for RegularSpatial

I think this is fine for now. Without an inheritance mechanism for docstrings, its a bit of a mess either way.

> Should I be concerned about the appveyor failures?

No. I'm not sure what's wrong with it, but it's definitely not your fault :)

PR #528, opened 2015-04-02
MAINT: Fix errors with sklearn-0.16

rmcgibbo:
Copy sklearn.utils validation function into source tree.


kyleabeauchamp:
Need to update  `__all__` in `utils/validation.py`?  Otherwise LGTM once that is fixed and travis passes.

kyleabeauchamp:
+1, failure is timeout

PR #532, opened 2015-04-06
BUG: Fix scaling of counts matrix

rmcgibbo:
cc @mpharrigan

PR #540, opened 2015-04-08
Make counts behave sensibly

### LINKED ISSUE #522: Turning off ergodic trimming gives different results
```python
        if self.ergodic_cutoff >= 1:
            # step 2. restrict the counts to the maximal strongly ergodic
            # subgraph
            self.countsmat_, mapping2 = _strongly_connected_subgraph(
                self.lag_time * raw_counts, self.ergodic_cutoff, self.verbose)
            self.mapping_ = _dict_compose(mapping, mapping2)
        else:
            # no ergodic trimming.
            self.countsmat_ = raw_counts
            self.mapping_ = mapping
```

case 1: multiply by lag time
case 2: will not get multiplied by lag time

makes a difference when using `prior_counts`

mpharrigan:
w.r.t. lag time, sliding window, and ergodic trimming

supercedes #532 and fixes #522 

Changes the behavior of `ergodic_cutoff` when `lag_time > 1` and `sliding_window == True`, but changes default behavior to be the same

### DIFF
--- a/msmbuilder/commands/fit.py
+++ b/msmbuilder/commands/fit.py
@@ -60,6 +60,12 @@ class MarkovStateModelCommand(FitCommand):
     _concrete = True
     _group = 'MSM'
 
+    def _ergodic_cutoff_type(self, erg):
+        if erg.lower() in ['on', 'off']:
+            return erg
+        else:
+            return float(erg)
+
 
 class BayesianMarkovStateModelCommand(FitCommand):
     klass = BayesianMarkovStateModel
--- a/msmbuilder/msm/core.py
+++ b/msmbuilder/msm/core.py
@@ -388,6 +388,11 @@ def _transition_counts(sequences, lag_time=1, sliding_window=True):
         other orderable objects.
     lag_time : int
         The time (index) delay for the counts.
+    sliding_window : bool
+        When lag_time > 1, considering *all*
+        ``N = lag_time`` strided sequences starting from index
+         0, 1, 2, ..., ``lag_time - 1``. The total, raw counts will
+         be divided by ``N``. When this is False, only start from index 0.
 
     Returns
     -------
@@ -425,7 +430,7 @@ def _transition_counts(sequences, lag_time=1, sliding_window=True):
     be counted. The mapping return value will not include the NaN or None.
     """
     if (not sliding_window) and lag_time > 1:
-        return  _transition_counts([X[::lag_time] for X in sequences], lag_time=1)
+        return _transition_counts([X[::lag_time] for X in sequences], lag_time=1)
 
     classes = np.unique(np.concatenate(sequences))
     contains_nan = (classes.dtype.kind == 'f') and np.any(np.isnan(classes))
@@ -473,7 +478,13 @@ def _transition_counts(sequences, lag_time=1, sliding_window=True):
         shape=(n_states, n_states))
     counts = counts + np.asarray(C.todense())
 
-    return counts / float(lag_time), mapping
+    # The following condition isn't strictly necessary: if sliding window
+    # is false, this function is called recursively at the beginning with
+    # strided trajectories and lag_time = 1. Dividing by 1 is nothing.
+    if sliding_window:
+        counts /= float(lag_time)
+
+    return counts, mapping
 
 
 def _dict_compose(dict1, dict2):
--- a/msmbuilder/msm/core.py
+++ b/msmbuilder/msm/core.py
@@ -389,7 +389,7 @@ def _transition_counts(sequences, lag_time=1, sliding_window=True):
     lag_time : int
         The time (index) delay for the counts.
     sliding_window : bool
-        When lag_time > 1, considering *all*
+        When lag_time > 1, consider *all*
         ``N = lag_time`` strided sequences starting from index
          0, 1, 2, ..., ``lag_time - 1``. The total, raw counts will
          be divided by ``N``. When this is False, only start from index 0.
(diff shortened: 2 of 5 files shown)


mpharrigan:
Commandline was failing with new options for `ergodic_cutoff`: in addition to a float, it should understand "on" and "off".

Fixed with latest commit. Involved changing some of the cmdline.py code

msmbuilder/msm/msm.py
```diff
@@ -191,11 +199,9 @@ def fit(self, sequences, y=None):
         return self
 
     def _fit_mle(self, counts):
-        if self.ergodic_cutoff < 1:
-            with warnings.catch_warnings(record=True):
```
mpharrigan: This warning was never being issued, btw

mpharrigan:
I don't know why it fails on 3.3. Seems unrelated

msmbuilder/msm/msm.py
```diff
@@ -54,24 +54,25 @@ class MarkovStateModel(BaseEstimator, _MappingTransformMixin, _SampleMSMMixin):
...
-    ergodic_cutoff : int, default=1
+    ergodic_cutoff : float or {'on', 'off'}, default='on'
         Only the maximal strongly ergodic subgraph of the data is used to build
         an MSM. Ergodicity is determined by ensuring that each state is
         accessible from each other state via one or more paths involving edges
         with a number of observed directed counts greater than or equal to
-        ``ergodic_cutoff``. Not that by setting ``ergodic_cutoff`` to 0, this
-        trimming is effectively turned off.
+        ``ergodic_cutoff``. Not that by setting ``ergodic_cutoff`` to 0 or
+        'off', this trimming is turned off.
```
rmcgibbo: what is the difference between `on` and a number?

mpharrigan: Are you asking or saying I should put it in the docs?

rmcgibbo:
I looked this over, and +1 to merge.

rmcgibbo:
@kyleabeauchamp do you want to look over this? (i've messed up this feature before... more eyes is probably better)

kyleabeauchamp:
Will do

msmbuilder/commands/fit.py
```diff
@@ -60,6 +60,12 @@ class MarkovStateModelCommand(FitCommand):
     _concrete = True
     _group = 'MSM'
 
+    def _ergodic_cutoff_type(self, erg):
```
kyleabeauchamp: Silly question, but can someone point me to where I can figure out how this member function gets used?  I presume it's described in the numpydocs?

kyleabeauchamp:
Another minor question: does the following block of code always work?  Note that it seems that classes can be *either* a list or a numpy array, but it is eventually broadcasted with `==`.  It seems to work on my machine, but I'm mosly curious.  

```
    if contains_nan:
        classes = classes[~np.isnan(classes)]
    if contains_none:
        classes = [c for c in classes if c is not None]

    n_states = len(classes)

    mapping = dict(zip(classes, range(n_states)))
    mapping_is_identity = np.all(classes == np.arange(n_states))
```

kyleabeauchamp:
One more question.  Is it just me or is line 452 redundant:

```
    counts = np.zeros((n_states, n_states), dtype=float)
```
I ask because it could possibly be doubling the memory usage of a dense N^2 array...

msmbuilder/msm/core.py
```diff
@@ -473,7 +478,13 @@ def _transition_counts(sequences, lag_time=1, sliding_window=True):
         shape=(n_states, n_states))
     counts = counts + np.asarray(C.todense())
 
-    return counts / float(lag_time), mapping
+    # The following condition isn't strictly necessary: if sliding window
```
kyleabeauchamp: I find it hard to understand what this comment is saying.  Is the code unnecessary?  If so, can we delete it?

msmbuilder/msm/msm.py
```diff
@@ -54,24 +54,26 @@ class MarkovStateModel(BaseEstimator, _MappingTransformMixin, _SampleMSMMixin):
...
         direct symmetrization of the expected number of counts.
-    ergodic_cutoff : int, default=1
+    ergodic_cutoff : float or {'on', 'off'}, default='on'
         Only the maximal strongly ergodic subgraph of the data is used to build
         an MSM. Ergodicity is determined by ensuring that each state is
         accessible from each other state via one or more paths involving edges
         with a number of observed directed counts greater than or equal to
-        ``ergodic_cutoff``. Not that by setting ``ergodic_cutoff`` to 0, this
-        trimming is effectively turned off.
+        ``ergodic_cutoff``. Not that by setting ``ergodic_cutoff`` to 0 or
```
kyleabeauchamp: "Not that" -> Just delete and say "By setting"

msmbuilder/commands/fit.py
mpharrigan: This is actually implemented in our `cmdline.py

mpharrigan: It's called when instantiating the class ("klass") which the command wraps

msmbuilder/msm/core.py
mpharrigan: The "if" could be removed, but it's more clear *why* we're dividing by lag time if it's kept

msmbuilder/msm/msm.py
mpharrigan: Good catch

kyleabeauchamp:
How about something like this instead:

```
    #  If sliding window is False, this function will be called recursively
    #  with lag_time = 1 and strided trajectories, which gives the
    #   correct number of counts.  If sliding window is True, [...]
    # strided trajectories and lag_time = 1. Dividing by 1 is nothing.
    counts /= float(lag_time)
```

kyleabeauchamp:
E.g. deleting the if statement and beefing up the comment

mpharrigan:
Ok, what do you think of latest commit

kyleabeauchamp:
+1

Issue #496, opened 2015-03-10
Model is *not* saved with the pickle protocol


mpharrigan:
The help text for the command lines say 

>Path to save fit model instance. This is the model,
>                        after parameterization with fit(), saved using the
>                        pickle protocol

But the models are *not* saved using the pickle protocol. They are saved using joblib, and if you try to `pickle.load` them, you get `Key error: Z`

cc @apeck12

Solutions:
 1. save using pickle format
 1. Change help text

I think we should do (1)

rmcgibbo:
I don't have strong feelings. (1) is fine.

kyleabeauchamp:
I'm happy with either, although I suppose we should double check that pickle isn't insanely slow for tICA models with modest numbers of features.

mpharrigan:
Empirically, I pickle tica models all the time (with O(1000) features) with no problems, although I think py3.4 has a fancier pickle protocol

kyleabeauchamp:
Sounds fine to me.

mpharrigan:
Steps
- [ ] measure a couple filesizes to make sure it's not different
- [ ] change to saving pickle
- [ ] change to load as pickle, load as joblib if that fails (backwards compatible)

mpharrigan:
```python
data = np.random.rand(500, 1000)
tica = tICA()
tica.fit([data])
print(tica.timescales_) # force solve
```

what | size | time
-------|------|---------
joblib | 29 mb | 900 ms
pickle 4 | 31 mb | 291 ms
pickle 3 | 31 mb | 281 ms
pickle 2 | 44 mb | 488 ms

note: joblib does some limited compression

kyleabeauchamp:
Cool

rmcgibbo:
+1

PR #541, opened 2015-04-09
Use pickle to save output

mpharrigan:
cc #496


mpharrigan:
I don't know why python 3.3 failed. It looks unrelated??

rmcgibbo:
maybe add a test for the load/save cycle, and to test/ensure that the old joblib files continue to work?

mpharrigan:
kk

rmcgibbo:
Awesome.

PR #544, opened 2015-04-13
Added more backports of sklearn utilities

kyleabeauchamp:
This one is used in osprey, but I figured I'd dump it here with the other backports.


rmcgibbo:
-1. If a validation function is needed for osprey but not msmbuilder, we should put it there.

PR #547, opened 2015-04-13
Add deprecation note to featurizer commandline

kyleabeauchamp:
```
Example Command
---------------
$ msmb DihedralFeaturizer --trj './trajectories/*.h5' \
    --out dihedrals-withchi --types phi psi chi1

optional arguments:
  -h, --help            show this help message and exit
  --chunk CHUNK         Chunk size for loading trajectories using
                        mdtraj.iterload (default: 10000)
  --out OUT             Output path for featurized data. NOTE: THIS ARGUMENT
                        WILL BE SWAPPED WITH --transformed IN MSMB 3.4!!!
  --stride STRIDE       Load only every stride-th frame (default: 1)
  --top TOP             Path to topology file matching the trajectories
  --trjs TRJS           Glob pattern for trajectories
```

### DIFF
--- a/msmbuilder/commands/featurizer.py
+++ b/msmbuilder/commands/featurizer.py
@@ -27,7 +27,7 @@ class FeaturizerCommand(NumpydocClassCommand):
     out = argument(
         '--out', required=True, help="""Output path for featurized data.
         NOTE: THIS ARGUMENT WILL BE SWAPPED WITH --transformed IN MSMB 
-        3.4!!!
+        3.3!!!
         """, type=exttype('/'))
     stride = argument(
         '--stride', default=1, type=int,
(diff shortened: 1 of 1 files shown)


kyleabeauchamp:
Wait, should this say deprecated in 3.3, right?

mpharrigan:
Right. 

Can we also let it accept `--transformed` here so people can start updating their scripts?

kyleabeauchamp:
Fixed the first part.  Regarding `--transformed`, it seems like that would be a bit more work, as I would have to implement logic to control the two different input flags.  I guess my vote is to just print out a warning for the present release.


```
Example Command
---------------
$ msmb DihedralFeaturizer --trj './trajectories/*.h5' \
    --out dihedrals-withchi --types phi psi chi1

optional arguments:
  -h, --help            show this help message and exit
  --chunk CHUNK         Chunk size for loading trajectories using
                        mdtraj.iterload (default: 10000)
  --out OUT             Output path for featurized data. NOTE: THIS ARGUMENT
                        WILL BE SWAPPED WITH --transformed IN MSMB 3.3!!!
  --stride STRIDE       Load only every stride-th frame (default: 1)
  --top TOP             Path to topology file matching the trajectories
  --trjs TRJS           Glob pattern for trajectories
```

PR #550, opened 2015-04-14
MAINT: Fix script for updating version dropdown

rmcgibbo:
In the 3.2.0 release, the update of the versions.json file that controls the dropdown box didn't work because of an SSL issue in the boto s3 library. See https://travis-ci.org/msmbuilder/msmbuilder/jobs/58469800

For pushing the docs, we changed a while ago from using boto to using the `s3cmd` command line tool. This does the same for the logic that updates the version.json dropdown.

### DIFF
--- a/devtools/ci/update-versions.py
+++ b/devtools/ci/update-versions.py
@@ -1,34 +1,47 @@
 from __future__ import print_function
 import os
-import boto
-from boto.s3.key import Key
-import msmbuilder.version
+import pip
+import json
+from tempfile import NamedTemporaryFile
+import subprocess
+from msmbuilder import version
+from six.moves.urllib.request import urlopen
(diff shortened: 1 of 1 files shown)


devtools/ci/update-versions.py
```diff
@@ -1,34 +1,47 @@
...
-    root = 'doc/_build'
 
-    versions = json.load(urllib2.urlopen('http://www.msmbuilder.org/versions.json'))
+versions = json.load(urlopen(URL))
 
-    # new release so all the others are now old
-    for i in xrange(len(versions)):
-        versions[i]['latest'] = False
+# new release so all the others are now old
+for i in xrange(len(versions)):
```
mpharrigan: change to `range` in case we ever switch to running things on 3.x?

rmcgibbo: done

PR #558, opened 2015-04-20
BUG: Fix bug in update-versions.py

rmcgibbo:


rmcgibbo:
Need to also set the `url` attribute for each version.

rmcgibbo:
Might also need to be applied to mdtraj -- I probably introduced the same bug there?

Issue #402, opened 2014-12-09
Unify GaussianHMM and VonMisesHMM.


rmcgibbo:
Another place where we could potentially get some help from @peastman is in rearchitecting the HMM code. Currently, we have two HMM classes that differer mathematically in a relatively small way -- they use a different emission distribution (Gaussian vs. Von Mises). But the code is 100% different. The Von mises model also uses a base class from `sklearn` which has been deprecated -- so that's not good.

The Gaussian model is able to use multiple threads in parallel, which is very nice, but it's not very extensible. The Von mises code uses a nicer subclassing design to avoid having the "von mises" part of the code reimplement functionality which is shared with other models. But that is all in python, and with the way the GIL works, it prevents the code from working on each of the sequence in parallel, which is the natural point of parallelism to decompose of the problem for multiple threads.

The most rational solution is to port everything into C++, because then you can do both parallelism and subclassing correctly. I tried this at one point, but I basically got stuck, because I don't understand C++ OO very well. https://github.com/msmbuilder/msmbuilder/pull/98/files

mpharrigan:
This has been done

Issue #162, opened 2014-05-15
`n_features` parameter in GHMM


kyleabeauchamp:
Shouldn't GHMM be able to figure out the correct value of `n_features`?  IMHO that would be more consistent with most other sklearn models.

PR #570, opened 2015-05-04
Rewrote GaussianFusionHMM

peastman:
This is a complete rewrite.  The goal of this project is to create a unified architecture for all HMM classes, thus eliminating code duplication, improving consistency, and allowing all of them to use optimized C++ implementations.  See #402.  The new GaussianFusionHMM is *almost* a drop-in replacement for the old one.  There are a few differences:

- The following constructor arguments have been removed: transmat_prior, params, init_params, precision.
- The argument n_em_iter has been renamed to n_iter to make GaussianFusionHMM and VonMisesHMM consistent with each other.
- A bunch of attributes that were previously read/write are now read only (means_, vars_, etc.).  These are things that get calculated, not things the user is supposed to modify.

Next steps include optimizing performance (vectorization and threading), and rewriting VonMisesHMM based on the same architecture.


mpharrigan:
Should we take this opportunity to change the namespace from mixtape to msmbuilder

rmcgibbo:
This looks great! I made a small benchmark that I'll send you. A couple minor comments:
1. Maybe we should rename it `GaussianHMM` instead of `GaussianFusionHMM`, since the L1 feature is optional.
2. Having `n_features` as a constructor argument is really a wart from the old implementation. Can we get rid of that? (#162)

peastman:
I'll make all three of those changes after group meeting.

peastman:
Can anyone tell what this error means?

```
======================================================================
ERROR: msmbuilder.tests.test_workflows.test_workflow_2
Traceback (most recent call last):
  File "/home/travis/miniconda/envs/_test/lib/python2.7/site-packages/nose/case.py", line 197, in runTest
    self.test(*self.arg)
  File "/home/travis/miniconda/envs/_test/lib/python2.7/site-packages/msmbuilder-3.3.0.dev0-py2.7-linux-x86_64.egg/msmbuilder/tests/test_workflows.py", line 32, in test_workflow_2
    subprocess.check_call(line.split())
  File "/home/travis/miniconda/envs/_test/lib/python2.7/subprocess.py", line 540, in check_call
    raise CalledProcessError(retcode, cmd)
CalledProcessError: Command '['msmb', 'GaussianFusionHMM', '--inp', 'atom_pairs_tica.h5', '--out', 'hmm.pkl', '--n_states', '2', '--n_features', '4']' returned non-zero exit status 2
```

rmcgibbo:
The `test_workflow` tests tries to run through a series of shell commands, emulating a user's actions with the msmbuilder command line API. That error means it tried running the shell command

```
$ msmb GaussianFusionHMM --inp atom_pairs_tica.h5 --out hmm.pkl --n_states 2 --n_features 4
```

and got a bad exit status.

peastman:
Here's what I get on my computer:

```
$ ~/anaconda/bin/msmb GaussianHMM --inp atom_pairs_tica.h5 --out hmm.pkl --n_states 2
usage: msmb [-h] [-v]  ...
msmb: error: unrecognized arguments: --n_states 2
```

What's causing that error?

rmcgibbo:
Not sure. I'm looking now.

rmcgibbo:
oh, grrr....

rmcgibbo:
This seems to fix the issue:

```
diff --git a/msmbuilder/cmdline.py b/msmbuilder/cmdline.py
index c4d18cf..a2b7142 100644
--- a/msmbuilder/cmdline.py
+++ b/msmbuilder/cmdline.py
@@ -313,6 +313,16 @@ class NumpydocClassCommand(Command):
         # mapping from the name of the argument to the helptext
         helptext = {d[0]: ' '.join(d[2]) for d in doc['Parameters']}
 
+        if not inspect.isfunction(cls.klass.__init__):
+            # inspect.getargspec fails for classes defined in cython.
+            # it's not clear exactly how to detect such classes,
+            # http://stackoverflow.com/q/30041371/1079728,
+            # but one thing that seems to work is looking at isfunction on the
+            # __init__, which, in the instances I checked, returns False for
+            # cython classes and True for most python classes. This is probably
+            # not the optimal solution...
+            args = list(helptext.keys())
+
         # mapping from the name of the argument to the type
         typemap = {d[0]: d[1].replace(',', ' ').split() for d in doc['Parameters']}
```

peastman:
Here's what I get with that addition:

```
$ ~/anaconda/bin/msmb GaussianHMM --inp atom_pairs_tica.h5 --out hmm.pkl --n_states 2
usage: msmb GaussianHMM [-h] -i INP -o OUT --n_init N_INIT --fusion_prior
                        FUSION_PRIOR --random_state RANDOM_STATE --vars_weight
                        VARS_WEIGHT --n_states N_STATES --n_jobs N_JOBS
                        --n_iter N_ITER --n_hotstart N_HOTSTART --timing
                        --init_algo INIT_ALGO --thresh THRESH --vars_prior
                        VARS_PRIOR --n_lqa_iter N_LQA_ITER --reversible_type
                        REVERSIBLE_TYPE
msmb GaussianHMM: error: the following arguments are required: --n_init, --fusion_prior, --random_state, --vars_weight, --n_jobs, --n_iter, --n_hotstart, --timing/--no-timing, --init_algo, --thresh, --vars_prior, --n_lqa_iter, --reversible_type
```

rmcgibbo:
Right, that makes sense.

peastman:
This looks like it's doing what you need:

https://github.com/neithere/argh/blob/master/argh/compat.py

See getargspec_permissive().

rmcgibbo:
Does it work for you? I get

```
----> 1 getargspec_permissive(GaussianHMM.__init__)

/home/rmcgibbo/miniconda/envs/3.4.2/lib/python3.4/site-packages/argh/compat.py in getargspec_permissive(func)
     43     # functions though such have perfectly usable func_code, func_defaults.
     44     if not (hasattr(func, "func_code") and hasattr(func, "func_defaults")):
---> 45         raise TypeError('{!r} missing func_code or func_defaults'.format(func))
     46 
     47     args, varargs, varkw = inspect.getargs(func.func_code)

TypeError: <slot wrapper '__init__' of 'msmbuilder.hmm.gaussian.GaussianHMM' objects> missing func_code or func_defaults
```

rmcgibbo:
(I'm working on py3, so it doesn't do anything special there)

rmcgibbo:
I think for the purpose of getting the docs built in mdtraj, we have cython classes like http://mdtraj.org/1.3.0/api/generated/mdtraj.formats.BINPOSTrajectoryFile.html#mdtraj.formats.BINPOSTrajectoryFile, and their docstring needs to be parsed to determine the constructor signature and parameters, and sphinx uses a trick where you need to write the signature on the first line of the docstring of the function https://github.com/mdtraj/mdtraj/blob/master/mdtraj/formats/binpos/binpos.pyx#L61.

so maybe we'll need to do something like that here.

rmcgibbo:
New idea: cython's embedsignatures option might work.... http://docs.cython.org/src/reference/compilation.html#compiler-directives

rmcgibbo:
Okay, that works to embed the signature. There's still some work to make a "fixed" version of inspect.getsignature and the numpydoc routines to use this properly...

rmcgibbo:
I filed a PR into your branch to fix the command line issue.

peastman:
Thanks!  I just merged it.  Let's see if the test passes now.

peastman:
It passed on Python 3.3 and 3.4, but failed on 2.7.

rmcgibbo:
The py2 error is from pickling.

```
Traceback (most recent call last):
  File "/home/rmcgibbo/miniconda/bin/msmb", line 9, in <module>
    load_entry_point('msmbuilder==3.3.0.dev0', 'console_scripts', 'msmb')()
  File "/home/rmcgibbo/projects/msmbuilder/msmbuilder/scripts/msmb.py", line 28, in main
    app.start()
  File "/home/rmcgibbo/projects/msmbuilder/msmbuilder/cmdline.py", line 435, in start
    instance.start()
  File "/home/rmcgibbo/projects/msmbuilder/msmbuilder/commands/fit.py", line 43, in start
    verbosedump(self.instance, self.out)
  File "/home/rmcgibbo/projects/msmbuilder/msmbuilder/utils/io.py", line 72, in verbosedump
    dump(value, fn, compress=compress)
  File "/home/rmcgibbo/projects/msmbuilder/msmbuilder/utils/io.py", line 41, in dump
    pickle.dump(value, f)
  File "/home/rmcgibbo/miniconda/lib/python2.7/pickle.py", line 1370, in dump
    Pickler(file, protocol).dump(obj)
  File "/home/rmcgibbo/miniconda/lib/python2.7/pickle.py", line 224, in dump
    self.save(obj)
  File "/home/rmcgibbo/miniconda/lib/python2.7/pickle.py", line 306, in save
    rv = reduce(self.proto)
  File "/home/rmcgibbo/miniconda/lib/python2.7/copy_reg.py", line 70, in _reduce_ex
    raise TypeError, "can't pickle %s objects" % base.__name__
TypeError: can't pickle GaussianHMM objects
Closing remaining open files:atom_pairs_tica.h5...done
```

rmcgibbo:
https://github.com/cython/cython/wiki/FAQ#how-do-i-pickle-cdef-classes

peastman:
Why does it work with Python 3?

rmcgibbo:
I'm not sure. I added the line `pickle.loads(pickle.dumps(model))` to `tests/test_ghmm.py:test_3` to get a proper traceback in the regular tests though.

peastman:
Ok, I added pickle support.  I verified that on Python 3, summarize() produces identical output for the original and unpickled objects, so it seems to be working correctly.  Let's see how it does on Python 2.

peastman:
Yay!  It worked!

Don't merge quite yet though.  I think I've found a bug in the accumulation of transition counts.

rmcgibbo:
On Windows VS9, it looks like `_USE_MATH_DEFINES` is necessary to get it to compile.
```
diff --git a/msmbuilder/hmm/src/GaussianHMMFitter.cpp b/msmbuilder/hmm/src/GaussianHMMFitter.cpp
index 79e3326..1cc8b24 100644
--- a/msmbuilder/hmm/src/GaussianHMMFitter.cpp
+++ b/msmbuilder/hmm/src/GaussianHMMFitter.cpp
@@ -2,6 +2,7 @@
 #ifndef DL_IMPORT
   #define DL_IMPORT(t) t
 #endif
+#define _USE_MATH_DEFINES
 #include "GaussianHMMFitter.h"
 #include "gaussian.h"
 #include <cmath>
```

peastman:
On Windows it now works in 32 bit mode, but fails in 64 bit mode:

```
Error: Unsatisfiable package specifications.
...
Hint: msmbuilder-dev 3.3.0.dev0 np19py27_0 has a conflict with the remaining packages
```

I suspect that has nothing to do with this PR?

rmcgibbo:
Yeah. Maybe @mpharrigan can help with that, but IMO it's a separate issue.

PR #572, opened 2015-05-07
Optimizations to Gaussian HMM

peastman:
This includes optimizations to computation and memory access in the two most expensive routines (compute_log_likelihood and accumulate_sufficient_statistics).  It also parallelizes most of the E-step across trajectories.

At this point, the majority of the time in the benchmark I'm using is spent in _init() and _do_mstep().  The former could presumably be made faster by changing the options for k-means, such as n_hotstart.  The most expensive part of the latter is the calls to linalg.solve().  I'm not sure if there's much I can do about that.


rmcgibbo:
Sweet. I tested on my machine (Ubuntu 14.04), and it passes as well.

PR #576, opened 2015-05-08
Tidy CI and quiet it down

mpharrigan:


mpharrigan:
includes the same fix as #575

Issue #574, opened 2015-05-07
Get Appveyor working


mpharrigan:
The reason why it's not straightforward is they only have the "express" versions of vc++2010 and 2008, which don't have normal x64 support

rmcgibbo:
Ideas:
 - Maybe we find another python project that is successfully using windows x64 on appveyor?
 - Maybe we can ask the people at Appveyor to install the right compilers?

rmcgibbo:
It looks to me like https://github.com/ogrisel/python-appveyor-demo is getting x64 working

mpharrigan:
re (2): I don't think they'd be very keen

http://help.appveyor.com/discussions/problems/466-no-support-for-visual-studio-c-2008-and-2010
^^ very relevant

http://poquitopicante.blogspot.com/2014/04/fix-vcvarsallbat-to-install-python-27.html
^^ I think a potential fix

I'm trying to reproduce the problem on a VM because it takes > 30 minutes for an appveyor build to even start

mpharrigan:
```
C:\miniconda3_x64\conda-bld\work>call "C:\Program Files (x86)\Microsoft Visual Studio 10.0\VC\vcvarsall.bat" amd64 
The specified configuration type is missing.  The tools for the
configuration might not be installed.
```

The relevant error message

mpharrigan:
I think the express versions of visual studio don't put the `vcvarsallxx.bat` files in the same place

rmcgibbo:
Do you have any idea of ogrisel's example is applicable?

mpharrigan:
Yeah that would probably work. I think that's essentially what we have now. It looks overengineered

http://stackoverflow.com/a/13751649

suggests that just finding the right vcvarsall can work

rmcgibbo:
Is the project status on https://ci.appveyor.com/project/rmcgibbo/msmbuilder/history accurate? It suggests to me that only py27-x64 is broken, and the error message is `Error: Unsatisfiable package specifications`.

mpharrigan:
Yeah it looks like the build actually is working (although tests aren't being run due to conda bug)

rmcgibbo:
Is that the conda bug you fixed earlier, or something different?

mpharrigan:
I fixed it upstream, and provided a work around in a branch that I haven't pr-ed yet.

The build's probably working now that I pushed all the dependency packages

rmcgibbo:
[...]

PR #577, opened 2015-05-08
Update test_libdistance.py

rmcgibbo:


rmcgibbo:
The intention here is to fix the 64-bit windows test failures https://github.com/msmbuilder/msmbuilder/issues/574#issuecomment-100334880

PR #580, opened 2015-05-09
ENH: Increase numerical stability of ratematrix

rmcgibbo:


kyleabeauchamp:
Seems reasonable to me, although it might be good to put a citation or note on where the small x expansion comes from.

PR #583, opened 2015-05-12
Rewrote VonMisesHMM

peastman:
This is a complete rewrite using the same architecture as the new GaussianHMM.  As with that one, this is mostly a drop-in replacement.  The most important difference is that it now supports the n_init argument to build several models and keep the best one.  For consistency with GaussianHMM, it defaults to 10, so if you run this with default arguments you should expect it to be slower than before.  That's because it's doing 10 times as much work!  We should consider what is really the best default for it.


msultan:
Out of curiosity whats the difference between the 10 models? Is it the location of the starting states? Since iirc, the EM algorithm always reaches a local maximum on the likelihood surface.

peastman:
Correct.  It initializes by doing a k-means, and that in turn is initialized with different random assignments.

peastman:
I don't know how much different it makes though.  Ideally, k-means would produce identical results independent of its starting state.  It doesn't, but if they're close enough, then the resulting HMMs will be even closer.  We need to study that.

kyleabeauchamp:
I recall once having an open issue somewhere on the tracker that discussed different initialization schemes, including KMeans and GMM.  I vaguely recall seeing differences for some real use cases.  Also: https://github.com/msmbuilder/msmbuilder/blob/228d8c947317e4c8ffd0b5944be8a0a813e9467e/msmbuilder/hmm/gaussian.pyx#L87

peastman:
It looks to me like the Windows builds are failing trying to install prerequisites.

rmcgibbo:
It's actually failing at

```
msmbuilder/msm\_ratematrix.cpp(3230) : error C2668: 'sqrt' : ambiguous call to overloaded function 
        C:\Program Files (x86)\Common Files\Microsoft\Visual C++ for Python\9.0\VC\Include\math.h(581): could be 'long double sqrt(long double)' 
        C:\Program Files (x86)\Common Files\Microsoft\Visual C++ for Python\9.0\VC\Include\math.h(533): or       'float sqrt(float)' 
        C:\Program Files (x86)\Common Files\Microsoft\Visual C++ for Python\9.0\VC\Include\math.h(128): or       'double sqrt(double)' 
        while trying to match the argument list '(int)'
```

rmcgibbo:
I tried to track this down on the other branch, and couldn't figure it out. I mean, I fixed that error just fine by casting to double, but I don't know _why_ it just started to occur. Maybe there was a new version of cython or something.

PR #584, opened 2015-05-12
Optimization to initializing GaussianHMM

peastman:
I replaced KMeans with the much faster MiniBatchKMeans.  I also eliminated the n_jobs argument, since MiniBatchKMeans doesn't have any corresponding argument.


rmcgibbo:
+1

peastman:
There seem to be several different causes of failures here.  The Windows builds still fail with a compilation error in msmbuilder/msm\_ratematrix.cpp.  For travis, the Python 3.4 build passed.  The Python 3.3 build had a test failure in msmbuilder.tests.test_libdistance.test_pdist_rmsd_2.  And the Python 2.7 build had a test failure that actually was related to GaussianHMM, though I think the problem is probably the test case, not the HMM code.  I'll try to make it more robust.

Issue #448, opened 2015-01-16
[FR] Map features back to atom indices


msultan:
I think that the DihedralFeaturizer class should include a method that maps each of the calculated feature to the residue index(using maybe the CA or something). I was thinking something like the following 
```
dihedral_feat = DihedralFeaturizer(["phi","psi","chi1"])
d_features = dihedral_feat.partial_transform(trj)

dihedral_feat.mapping(40)
=23
```

I think it would make the analysis a lot easier. Probably a lot of the other featurizers could benefit as well.

msultan:
I can work on this if people are busy but I would like some input on how to best implement it.

kyleabeauchamp:
I think having a more full-featured feature exploration API is a good idea, let's discuss.

msultan:
Ideally, it should be able to take in a featurizer or a list of featurizer and fit/transform using all/subset of them and be aware of what features went it and their atomic mapping. 

Personally, I would prefer each featurizer have a mapping method from the column index to atom indices/residue indices used. For RMSD and contacts, this could be a list. 

Also, scikit learn has the feature_extraction but its very awkward imho. 
http://scikit-learn.org/stable/modules/feature_extraction.html

msultan:
It would be nice/though not necessary, if the feature explorer perhaps took in any of the decomposition method like pca/tica/s-tica and returned the top m features in the top n components.

kyleabeauchamp:
So I also have some incomplete code for selecting subsets of features: https://github.com/msmbuilder/msmbuilder/blob/master/msmbuilder/featurizer/subset_featurizer.py

I'd love to figure out what to do with that incomplete piece of code.  

I also had some code that used monte carlo to attempt to find the "best" subset of an existing group of featurizers.  I never had time to finish it though, and now you and Robert both have competing approaches as well.  

IMHO, the most pressing questions are:

1.  What do we need in terms of feature extraction / choosing subsets of features
2.  How do we best keep track of feature metadata
3.  How do we build a nice way to get human-readable visualizations / representations of geometric features-e.g. for writing papers on what the features *mean*

msultan:
1). Imo, i think its still an open question which feature selection method is "best" within a biological context. There are some problems with each(including mine). Which is why, I think we should support a few of them. 
2). Not so sure. 
[...]

PR #586, opened 2015-05-13
added atom indices to dihedral featurizer and kappa featurizer 

msultan:
Changed the dihedral and kappa featurizers to save atom indices used. This somewhat implements  https://github.com/msmbuilder/msmbuilder/issues/448 . I welcome any other ways in which this might be done.

### DIFF
--- a/msmbuilder/featurizer/featurizer.py
+++ b/msmbuilder/featurizer/featurizer.py
@@ -216,6 +216,7 @@ class AtomPairsFeaturizer(Featurizer):
         # TODO: We might want to implement more error checking here. Or during
         # featurize(). E.g. are the pair_indices supplied valid?
         self.pair_indices = pair_indices
+        self.atom_indices = pair_indices
         self.n_features = len(self.pair_indices)
         self.periodic = periodic
         self.exponent = exponent
@@ -267,6 +268,8 @@ class DihedralFeaturizer(Featurizer):
             types = [types]
         self.types = list(types)  # force a copy
         self.sincos = sincos
+        self.wrote_atom_indices = False
+        self.atom_indices = []
 
         known = {'phi', 'psi', 'omega', 'chi1', 'chi2', 'chi3', 'chi4'}
         if not set(types).issubset(known):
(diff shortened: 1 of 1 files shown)


kyleabeauchamp:
Yeah, the only ideas I had was either this or creating a separate object that duplicates the code in DihedralFeaturizer but without actually outputting the trajectories.  Something like below, which I'm only including to keep a record of this.

```
def describe_features(self, traj):
    """Return a Pandas Dataframe describing the features."""
    x = []
    for a in self.types:
        func = getattr(md, 'compute_%s' % a)
        aind, y = func(traj)
        n = len(aind)
        
        aind = aind[:, 1]
        
        resnames = [traj.top.atom(i).residue.name for i in aind]
        resSeq = [traj.top.atom(i).residue.resSeq for i in aind]
        resid = [traj.top.atom(i).residue.index for i in aind]
        bigclass = ["dihedral"] * n
        smallclass = [a] * n
        
        if self.sincos:
            #x.extend([np.sin(y), np.cos(y)])
            resnames = resnames * 2
            resSeq = resSeq * 2
            resid = resid * 2
            trig = (["sin"] * n) + (["cos"] * n)
            bigclass = bigclass * 2
            smallclass = smallclass * 2
        else:
            trig = ["nosincos"] * n
        
        for i in range(len(resnames)):
            d_i = dict(resname=resnames[i], resSeq=resSeq[i], resid=resid[i], trig=trig[i], bigclass=bigclass[i], smallclass=smallclass[i])
            x.append(d_i)
    
    return x

msultan:
I think pandas df might be a better idea than my array of arrays though I think saving residue information is problematic since many features involve multiple residues. I was thinking we save the information at the atom level and then combine that to get residue level information.

msultan:
So I had a conversation with @mpharrigan  and @rmcgibbo about this offline. Matt thinks we should have a separate method which is topology dependent so it makes sense to use something akin to what kyle suggested above.

msultan:
Gonna close this and continue conversation in https://github.com/msmbuilder/msmbuilder/issues/448 since this requires a little more thought

PR #588, opened 2015-05-14
Fix windows compiler issues

mpharrigan:
The ratematrix code has decided to be compiled as c++ now, and this was leading to some compiler errors on windows


rmcgibbo:
Sweet. This fixes the compilation errors.

PR #599, opened 2015-06-09
Wrap KDTree for finding nearest points

mpharrigan:


mpharrigan:
Should I hack the docs from scipy's kdtree a la clustering docs

mpharrigan:
Docstrings added

mpharrigan:
bump :)

PR #614, opened 2015-07-22
Documentation updates 1

mpharrigan:


rmcgibbo:
should we merge this?

mpharrigan:
No, I'll pick it up and make it mergeable. 

iirc I think it breaks some things

mpharrigan:
Can we make the default msmbuilder.org direct to the latest full release so I can e.g. upload an example notebook that uses features from the development version

rmcgibbo:
yeah, we can make it redirect wherever.

mpharrigan:
Can you do that? I think the relevant code is in msmb_theme but I can't make heads or tails of it

rmcgibbo:
What URL do you want msmbuilder.org to redirect to exactly?

mpharrigan:
http://msmbuilder.org/3.3.0/ or whatever is marked as `{"latest": true}` in http://msmbuilder.org/versions.json

rmcgibbo:
So currently there's a file that you can get `wget http://msmbuilder.org/index.html` that redirects to `/latest`, which is where the development docs are.

rmcgibbo:
Okay, I'll change the index.html file to

```
<html><head>
<!-- redirect to /latest after 3 seconds, in case something fails or
    the user has javascript disabled -->
<meta http-equiv="refresh" content="3;URL='/latest'"/>

<script type="text/javascript">
// download the versions.json file and redirect to the URL associated
// with the entry that has latest=true
var xhr = new XMLHttpRequest();
xhr.open('GET', encodeURI('/versions.json'));
xhr.onload = function() {
    if (xhr.status === 200) {
        var versions = JSON.parse(xhr.responseText);
        for (var key in versions) {
            if (versions[key]['latest']) {
                var latest = versions[key];                
                window.location.assign(latest['url']);
            }
        }
    }
}
xhr.send();
</script>
</head></html>
```

rmcgibbo:
Done.

mpharrigan:
Going to msmbuilder.org/index.html now works, but msmbuilder.org still directs you to latest/

mpharrigan:
Nevermind. It looks like it was using a cached version

mpharrigan:
Ok, this is in a workable state. It requires the use of https://github.com/msmbuilder/msmb_theme/pull/6

Maybe we can pin mdtraj's doc-building to the current version of msmb_theme, merge this, and continue to work on documentation content.

mpharrigan:
Woo! it builds on python 3.4

I'm going to remove python 3.3 (no nbconvert package)

I'm investigating why the notebook runner doesn't work on python 2.7

rmcgibbo:
all the deps should be available for py35 now too.

mpharrigan:
It was because the notebook had a specification for a python3 kernel. Changing it to "python" seems to do the trick locally

PR #625, opened 2015-08-11
Use Travis-CI container infrastructure

mpharrigan:


mpharrigan:
ratematrix failures seem unrelated

Issue #628, opened 2015-08-12
Fix tests on Python 3.3


mpharrigan:
These recently started breaking reliably

mpharrigan:
I can't reproduce this locally

mpharrigan:
They seem to be the same failures as we were seeing on appveyor #594 and seemed to have started when we switched to the "container" infrastructure on travis

Issue #594, opened 2015-05-28
Python 3 test failures on windows


mpharrigan:
There are still a couple failures on appveyor for python 3
```
======================================================================
ERROR: msmbuilder.tests.test_msm_uncertainty.test_1
----------------------------------------------------------------------
Traceback (most recent call last):
  File "C:\Python34_32\envs\_test\lib\site-packages\nose\case.py", line 198, in runTest
    self.test(*self.arg)
  File "c:\python34_32\envs\_test\lib\site-packages\msmbuilder-3.3.0.dev0-py3.4-win32.egg\msmbuilder\tests\test_msm_uncertainty.py", line 61, in test_1
    print(model2.uncertainty_timescales())
  File "c:\python34_32\envs\_test\lib\site-packages\msmbuilder-3.3.0.dev0-py3.4-win32.egg\msmbuilder\msm\ratematrix.py", line 263, in uncertainty_timescales
    self._build_information()
  File "c:\python34_32\envs\_test\lib\site-packages\msmbuilder-3.3.0.dev0-py3.4-win32.egg\msmbuilder\msm\ratematrix.py", line 300, in _build_information
    self.information_[np.ix_(inds, inds)] = scipy.linalg.pinv(-hessian)
  File "C:\Python34_32\envs\_test\lib\site-packages\scipy\linalg\basic.py", line 602, in pinv
    a = np.asarray_chkfinite(a)
  File "C:\Python34_32\envs\_test\lib\site-packages\numpy\lib\function_base.py", line 613, in asarray_chkfinite
    "array must not contain infs or NaNs")
nose.proxy.ValueError: array must not contain infs or NaNs
-------------------- >> begin captured stdout << ---------------------
loading "C:\Users\appveyor\msmbuilder_data\doublewell\version-1_random-state-0.pkl"...
MSM uncertainty timescales:
[ 108.53833321    1.53644089    0.99740744    0.69777323    0.3952434
    0.35207525    0.29733261    0.20266253    0.13963238]
ContinuousTimeMSM uncertainty timescales:
[ 105.39363654    1.51931341    0.98284022    0.69910175    0.39404041
    0.36077648    0.30040502    0.20571821    0.14115993]
 
MSM uncertainty timescales:
[ 101.02181988    1.10618418    1.79733039    0.70170151    0.39851563
    0.36410131    0.1687596     0.20884279    0.2933363 ]
ContinuousTimeMSM uncertainty timescales:
[...]

PR #631, opened 2015-08-12
Initialize x, y in muller cython code

mpharrigan:
`x` and `y` were declared but not initialized to a value (should be `x0[0]` and `x0[1]` function parameters)

I assume most of the time these were getting set to zero. On certain platforms on travis and appveyor they must have been set to garbage. 

This PR also tidies some of the testing code. The meat of the bugfix is in 15b0ee0


mpharrigan:
fixes #628 
fixes #594

PR #642, opened 2015-08-27
Change default n_timescales to n_states - 1

### LINKED ISSUE #603: Wrong lumping results when using PCCA/PCCA+ with more than 10 macrostates
I believe this bug is due to that the number of eigenvectors calculated by the class MarkovStateModel() is limited by the n_timescales which is default to 10. In class PCCA/PCCA+() the n_macrostates will not update the n_timescales. So if I use more than 10 macro states, the results will still give me only 10 macro states.
Could you check on that please? Thanks.

mpharrigan:
The default for the number of timescales was 10. This caused problems
with implied_timescales (needed to pass n_timescales twice) and with
PCCA(+) (would fail for > 10 macrostates).

Fixes #603

### DIFF
--- a/msmbuilder/msm/msm.py
+++ b/msmbuilder/msm/msm.py
@@ -350,10 +350,8 @@ class MarkovStateModel(BaseEstimator, _MappingTransformMixin, _SampleMSMMixin):
                     self._left_eigenvectors,
                     self._right_eigenvectors)
 
-        n_timescales = self.n_timescales
-        if n_timescales is None:
-            n_timescales = self.n_states_ - 1
-        n_timescales = min(n_timescales, self.n_states_ - 1)
+        n_timescales = min(self.n_timescales if self.n_timescales is not None
+                           else self.n_states_ - 1, self.n_states_ - 1)
 
         k = n_timescales + 1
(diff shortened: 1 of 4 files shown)


mpharrigan:
Does anyone have any idea why this failed on py33?

rmcgibbo:
I assume you restarted it? It's passing now.

mpharrigan:
Yeah, I restarted it. We'll keep an eye on that test to see if it's too stochastic.

msmbuilder/msm/msm.py
```diff
@@ -581,9 +582,11 @@ def uncertainty_eigenvalues(self):
         if self.reversible_type is None:
             raise NotImplementedError('reversible_type must be "mle" or "transpose"')
 
-        n_timescales = min(self.n_timescales, self.n_states_ - 1)
+        n_timescales = self.n_timescales
         if n_timescales is None:
             n_timescales = self.n_states_ - 1
+        n_timescales = min(n_timescales, self.n_states_ - 1)
```
rmcgibbo: this logic (585-588) can probably be simplified down to 1 line.

mpharrigan: how?

PR #672, opened 2015-10-23
Documentation updates 2

mpharrigan:
Finalize deployment.


mpharrigan:
Note that the docs from #614 can be seen here http://msmbuilder.org/test-deploy/development/html/index.html

rmcgibbo:
looks good.

rmcgibbo:
what would be really sick is if we could connect the head maps with the mdtraj webgl protein visualizer in the tutorial.

mpharrigan:
Now here: http://msmbuilder.org/development/

PR #673, opened 2015-10-24
Docs: once over on content

mpharrigan:
Go through the pages, format rst, make sentences better


mpharrigan:
I'm going to merge this so we can see a rendered version. It isn't the default for when users type msmbuilder.org (anymore) so it should be fine

PR #676, opened 2015-10-28
test from brooke

brookee:
hi


mpharrigan:
good job submitting a pull request

PR #675, opened 2015-10-28
PR for describe_features method

msultan:
This PR implements the describe_features method for 3 of the most commonly used featurziers ```ContactFeaturizer, DihedralFeaturizer, & KappaAngleFeaturizer```. Most, if not all, of the code was inspired by Kyle's previous attempt to get this going. 

For the ```DihedralFeaturizer``` 
```python

In [13]: f=DihedralFeaturizer()

In [14]: df = pd.DataFrame(f.describe_features(t))

In [15]: df[:1]
Out[15]: 
            atomind  bigclass otherInfo  resSeq   resid     resname smallclass
0  [22, 24, 26, 34]  dihedral       sin  [1, 2]  [0, 1]  [LYS, ASP]        phi
```

For the ```ContactFeaturizer``` 
```python
In [16]: f = ContactFeaturizer()

In [17]: df = pd.DataFrame(f.describe_features(t))
 
In [18]: df[:1]
Out[18]: 
  atomind bigclass otherInfo  resSeq   resid     resname     smallclass
0     N/A      all      True  [1, 4]  [0, 3]  [LYS, TRP]  closest-heavy
```

For the ```KappaAngleFeaturizer```
```python
In [26]: f = KappaAngleFeaturizer()

In [27]: df = pd.DataFrame(f.describe_features(t))

In [28]: df[:1]
Out[28]: 
       atomind bigclass otherInfo     resSeq      resid          resname  \
0  [4, 38, 78]    angle       cos  [1, 3, 5]  [0, 2, 4]  [LYS, VAL, GLU]   

  smallclass  
0      kappa 
 ```

PR #677, opened 2015-11-04
Clean up describe_features

mpharrigan:
cc #675

### DIFF
--- a/msmbuilder/featurizer/featurizer.py
+++ b/msmbuilder/featurizer/featurizer.py
@@ -14,6 +14,7 @@ import sklearn.pipeline
 from sklearn.externals.joblib import Parallel, delayed
 from scipy.stats import vonmises as vm
 from msmbuilder import libdistance
+import itertools
 
 from ..base import BaseEstimator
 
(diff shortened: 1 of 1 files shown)


mpharrigan:
@msultan can you look at this, check it still works, think of a test we can add

msultan:
The ```set``` logic doesn't work if you have two or more of the same type of residues in a row and you are iterating over the atom indices. 

```python
In [34]: df[509:510]
Out[34]: 
                     atominds  featuregroup featurizer           resids  \
509  [3975, 3977, 3988, 3990]  nosincos-psi   Dihedral  set([248, 247])   

       resnames          resseqs  
509  set([GLU])  set([248, 249])  

```
Since we are changing this, would changing resids to residue_index make more sense?

mpharrigan:
Why do you want only unique resids and resseqs but all the resnames

msultan:
seems more consistent imho.

mpharrigan:
> Why do you want only unique resids and resseqs but all the resnames

This is the opposite of consistent. Would you be ok with returning a four-tuple for each relevant field?

msultan:
I thought about this more. I think the unique implementation makes more sense. If a user wants more detail, they can iterate over the atom indices then. 
Good to merge imo.

mpharrigan:
Wait, I think I just figured out what you may have actually been saying instead of what I thought you were saying. 

You want the four atom indices and info on the residues to which they belong. i.e. unique resids and for each unique resid, the associated resseq and resname. You could use this info to color residues in vmd or something

msultan:
Ya thought the unique resid/resSeq is already there so I think we are fine. I was worried about the resname but I think its okay to keep that unique as well for consistency. 
Also what do you think about adding ```chain id``` ? Might be useful in multiprotein systems or membrane proteins with missing loops.

msultan:
@mpharrigan do you wanna merge this and I can use the pattern to fix the ```describe_features``` for other featurizers too?

mpharrigan:
Let me fix it to do what you actually wanted it to do re: Number of resseq entries

mpharrigan:
@msultan can you check the latest version of this. It will find unique resids and give the resids, resnames, and resseq all as lists of length(number of unique residues)

cxhernandez:
also add a test for https://github.com/msmbuilder/msmbuilder/issues/739

msultan:
gonna do this tomorrow. sry for being awol all day, i was trying to find the damn h264 codec on my mac

msultan:
also, should we fix the other ```describe_feature``` attributes as well?

mpharrigan:
Can you do that?

PR #683, opened 2015-11-09
No longer push msmbuilder-dev

mpharrigan:
This has outlived its usefulness. No one should be using this conda package.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -43,9 +43,3 @@ deploy:
     on:
       branch: master
       condition: "$CONDA_PY = 3.4"
-
-  - provider: script
-    script: devtools/ci/deploy_conda.sh
-    skip_cleanup: true
-    on:
-      branch: master
--- a/devtools/ci/deploy_conda.sh
+++ b/devtools/ci/deploy_conda.sh
@@ -1,4 +1,0 @@
-#!/bin/bash
-
-conda install --yes anaconda-client
-anaconda -t $BINSTAR_TOKEN upload -u omnia --force `conda build devtools/conda-recipe --output`
--- a/devtools/conda-recipe/meta.yaml
+++ b/devtools/conda-recipe/meta.yaml
@@ -1,6 +1,6 @@
 package:
   name: msmbuilder-dev
-  version: !!str 3.4.0.dev0
+  version: "3.4.0.dev0"
 
 source:
   path: ../../


jchodera:
Why no dev packages?

mpharrigan:
We started pushing these in early days of msmbuilder3 when the apis of both this and mdtraj were changing rapidly enough that we needed the dev packages for testing. I can't imagine us adding a feature to either package that depends on the dev version of the other.

Users should be using the released versions. Developers should be comfortable installing from source, with the added benefit that they'll have some idea of what commit they're building/using

jchodera:
> We started pushing these in early days of msmbuilder3 when the apis of both this and mdtraj were changing rapidly enough that we needed the dev packages for testing. I can't imagine us adding a feature to either package that depends on the dev version of the other.

> Users should be using the released versions. Developers should be comfortable installing from source, with the added benefit that they'll have some idea of what commit they're building/using

There is a great amount of power in having all the `omnia` packages be uniform in the way they handle conda packaging.  We may want to build a tool that uses `msmbuilder` to do, say, automated MSM construction and want to make sure it works with both the current release version and in-development version of `msmbuilder`.  (We do this with OpenEye tools, for example!)  It is super helpful that both packages are available through conda so automated build/test systems can use the conda packages.

Can you leave the `openmm-dev` package?  I can see a lot of benefit from keeping this updated, and really no harm in continuing to build it.

rmcgibbo:
This isn't about OpenMM.

rmcgibbo:
+1 from me, as well.

jchodera:
> This isn't about OpenMM.

Sorry, I meant `msmbuilder-dev`.

We're working on tools that automatically build MSMs for kinases right now.  It would be great to be able to test them against both `msmbuilder` and `msmbuilder-dev`. 

The whole point of omnia is to have a "package ecosystem" where we can introduce dependencies between tools.  Testing is facilitated by having these packages built for both release and dev versions.  Why break this?

rmcgibbo:
- As developers, we want to have the freedom to break stuff in master between releases. We do not want to provide a guarantee of a rolling commitment that everything works -- we want to do that at release time. By providing a compiled version of the development head, we're compromising that and providing an implied warrant of fitness that is annoying, and holds back development.
- It's a perverse incentive against making frequent releases.
- `pip install https://github.com/msmbuilder/msmbuilder/archive/master.zip` is always one command away.

jchodera:
> `pip install https://github.com/msmbuilder/msmbuilder/archive/master.zip` 

Does this automatically resolve dependencies?

jchodera:
> As developers, we want to have the freedom to break stuff in master between releases. We do not want to provide a guarantee of a rolling commitment that everything works -- we want to do that at release time. By providing a compiled version of the development head, we're compromising that and providing an implied warrant of fitness that is annoying, and holds back development.

I'm trying to figure out what implications this has for how we specify build constraints for release packages.  

Suppose we have a tool that depends on `msmbuilder`, and I know it works with the API in `3.3.0`.  If I specify the constraint `msmbuilder >=3.3.0`, and the developers change the API in a new release (3.4.0 or 4.0.0), a new user downloading my tool will find that it doesn't work.  I won't have any idea that this is broken until either a user complains or I commit a new change to my project (in which case travis will fail if dependencies are specified the same way), but in any case, I won't know until after the release.

Ont he other hand, if I check against `msmbuilder-dev` in my travis builds, I am able to detect if an API changes in a way that breaks my code during the development process *before* the release if I make commits during this time.  That gives me time to adapt to API changes or modify the dependencies.  Isn't that desirable?

Really, maybe this just means we should *never* be specifying dependencies for releases as `package` or `package >=version`, but always `package ==version` or `package >=version <=version`.  Does that make sense?

rmcgibbo:
We're been pretty conservative about breaking the API from release to release, but that doesn't mean every individual commit along the way comes with the same guarantee.

jchodera:
Sure.

So what is the best way to handle the cases where the API might change but
we don't want downstream released packages to break?

What if we enforced across omnia that major version numbers MUST change if
the API changes?  Then we could limit release dependencies to the major
version number.

Issue #366, opened 2014-12-03
Hierarchical clustering needs clean up.


rmcgibbo:
Currently, we have a wrapper for `sklearn.clustering.Ward` and `LandmarkAgglomerative`. There are a number of problems.

 - The Ward class in sklearn is deprecated.
 - Neither of these classes use libdistance, which means they don't handle rmsd using the same pattern as KCenters, KMedoids, etc.
 - None of the models are exposed on the command line

We should make sure all of these classes have a transform method to assign labels for new datasets, even if the method for computing these labels is arguably not optimal. Without this capability, with GMRQ, the models are basically useless. `LandmarkAgglomerative` does this somewhat reasonably, IMO, using `POOLING_FUNCTIONS`, for `average`, `complete` and `single` linkage, but I'm not convinced we can't to something reasonable for Ward. Even, I think, using the `average` linkage pooling function with clusters that were generated using `Ward` would be reasonable, I think.

Also, the ability to assign labels to new datasets is _essential_ for production calculations on non-toy datasets, since all of these require O(N^2) memory. Naturally, this means that we're going to be doing the clustering on either a strided or randomly-sampled subset of the data, so then we're really going to want to assign state labels for the full dataset.

IMO, we should probably use `fastcluster` / `scipy.cluster.hierarchy`, and skip using `sklearn.cluster.AgglomerativeClustering`

mpharrigan:
@brookehus did you do this? Can this be closed?

brookehus:
Yup, all three bullet points have been addressed in #874.

Issue #280, opened 2014-11-12
sklearn Ward warning


kyleabeauchamp:
```
In [7]: ward = mixtape.cluster.Ward(n_clusters=10)
/home/kyleb/opt/lib/python2.7/site-packages/sklearn/cluster/hierarchical.py:763: DeprecationWarning: The Ward class is deprecated since 0.14 and will be removed in 0.17. Use the AgglomerativeClustering instead.
  "instead.", DeprecationWarning)
```

Probably good to deal with this at some point.

Also, I wonder if we should should try to expose fastcluster algorithms via an sklearn api?  This is mostly as a way to improve feature parity with MSMB2

PR #685, opened 2015-11-09
Remove Ward

mpharrigan:
`import msmbuilder.cluster` throws an exception with latest sklearn. They removed Ward


mpharrigan:
cc #280 #366

PR #687, opened 2015-11-09
Run tests on py3.5

mpharrigan:


mpharrigan:
I thought @rmcgibbo had all the omnia packages built for windows py35, but alas

rmcgibbo:
> I thought @rmcgibbo had all the omnia packages built for windows py35, but alas

I think this is fixed now.

PR #693, opened 2015-11-19
Rmse

brookehus:
superposefeaturizer now outputs an array with correct dimensions


cxhernandez:
You'll have to fix the test as well.

mpharrigan:
Can you describe what the previous behavior was, why it was incorrect, and what the fix does

msmbuilder/tests/test_commands.py
```diff
@@ -133,7 +133,7 @@ def test_superpose_featurizer():
                   data_home=get_data_home()))
         ds = dataset('distances')
         assert len(ds) == 10
-        assert ds[0].shape[1] == len(np.loadtxt('all.txt'))
+        assert ds.shape[1] == len(np.loadtxt('all.txt'))
```
mpharrigan: I'm pretty sure this is wrong. `dataset` objects don't have a shape attribute

msmbuilder/tests/test_ghmm.py
```diff
@@ -19,7 +19,7 @@ def test_1():
     indices = topology.select('symbol C or symbol O or symbol N')
     featurizer = SuperposeFeaturizer(indices, trajectories[0][0])
 
-    sequences = featurizer.transform(trajectories)
+    sequences = featurizer.partial_transform(trajectories)
```
mpharrigan: I think transform is (still) appropriate here

cxhernandez: I agree. What is `trajectories` though? A list of `mdtraj.Trajectory` objects?

msmbuilder/tests/test_commands.py
cxhernandez: This is my b. 

@brookehus: In hindsight, this should be:
```
assert ds[0].shape[0] == len(np.loadtxt('all.txt'))
```

mpharrigan:
I'm still confused what this pr is trying to achieve

PR #695, opened 2015-11-22
featurizer: rmsd's of frames to any number of reference frames

brookehus:
As discussed with @mpharrigan this pull request attempts to create a featurizer which inputs a trajectory to be analyzed ("traj") and a reference trajectory ("ref") and outputs the RMSD of each frame of traj with respect to each frame in ref. The intended output is a numpy array with n_rows = traj.n_frames and n_columns = ref.n_frames. I have also included tests for ref.n_frames = 1 and ref.n_frames = 2.

Question - I don't think my python in the featurizer code is elegant. I want to output an array as specified above but it seems to output a list that corresponds to a 3D array with one extra/meaningless dimension. Upon fixing this we must also fix the second test in tests/test_strucrmsdfeaturizer.py; see TODO.

cc: @rmcgibbo @cxhernandez

### DIFF
--- a/msmbuilder/featurizer/featurizer.py
+++ b/msmbuilder/featurizer/featurizer.py
@@ -16,7 +16,7 @@ import mdtraj as md
 from sklearn.base import TransformerMixin
 import sklearn.pipeline
 from sklearn.externals.joblib import Parallel, delayed
-import libdistance
+from msmbuilder import libdistance
 
 from ..base import BaseEstimator
 
--- a/msmbuilder/featurizer/featurizer.py
+++ b/msmbuilder/featurizer/featurizer.py
@@ -254,6 +254,8 @@ class StrucRMSDFeaturizer(Featurizer):
         result = np.array(x)
         return np.transpose(x)
 
+        # test 2
+
         # for i in range(0, self.reference_traj.n_frames):
         #     y = np.array([md.rmsd(traj, self.reference_traj, i)])
         #     x = np.append(x,y)
--- a/msmbuilder/featurizer/featurizer.py
+++ b/msmbuilder/featurizer/featurizer.py
@@ -16,6 +16,7 @@ import mdtraj as md
 from sklearn.base import TransformerMixin
 import sklearn.pipeline
 from sklearn.externals.joblib import Parallel, delayed
+import libdistance
 
 from ..base import BaseEstimator
 
@@ -249,7 +250,7 @@ class StrucRMSDFeaturizer(Featurizer):
         --------
         transform : simultaneously featurize a collection of MD trajectories
         """
-        result = cdist(traj, self.reference_traj, 'rmsd')
+        result = libdistance.cdist(traj, self.reference_traj, 'rmsd')
         return result
 
 
--- a/msmbuilder/featurizer/featurizer.py
+++ b/msmbuilder/featurizer/featurizer.py
@@ -253,6 +253,7 @@ class StrucRMSDFeaturizer(Featurizer):
             
         result = np.array(x)
         return np.transpose(x)
+        # test
         # for i in range(0, self.reference_traj.n_frames):
         #     y = np.array([md.rmsd(traj, self.reference_traj, i)])
         #     x = np.append(x,y)
--- a/msmbuilder/featurizer/featurizer.py
+++ b/msmbuilder/featurizer/featurizer.py
@@ -252,15 +252,7 @@ class StrucRMSDFeaturizer(Featurizer):
             x.append(y)
             
         result = np.array(x)
-        return np.transpose(x)
-
-        # test 2
-
-        # for i in range(0, self.reference_traj.n_frames):
-        #     y = np.array([md.rmsd(traj, self.reference_traj, i)])
-        #     x = np.append(x,y)
-
-        # return np.hstack(x)
+        return np.transpose(result)
 
 
 class AtomPairsFeaturizer(Featurizer):
--- a/msmbuilder/featurizer/featurizer.py
+++ b/msmbuilder/featurizer/featurizer.py
@@ -200,6 +200,60 @@ class SuperposeFeaturizer(Featurizer):
         return x
 
 
+
+class StrucRMSDFeaturizer(Featurizer):
+    """Featurizer based on overall RMSD to a set of reference structures.
+
+    Parameters
+    ----------
+    atom_indices : np.ndarray, shape=(n_atoms,), dtype=int
+        The indices of the atoms to superpose and compute the distances with
+    reference_traj : md.Trajectory
+        The reference conformation to superpose each frame with respect to
+        (only the first frame in reference_traj is used)
+    superpose_atom_indices : np.ndarray, shape=(n_atoms,), dtype=int
+        If not None, these atom_indices are used for the superposition
+    """
+
+    def __init__(self, atom_indices, reference_traj, superpose_atom_indices=None):
+        self.atom_indices = atom_indices
+        if superpose_atom_indices is None:
+            self.superpose_atom_indices = atom_indices
+        else:
+            self.superpose_atom_indices = superpose_atom_indices
+        self.reference_traj = reference_traj
+#        self.n_features = len(self.atom_indices)
+
+    def partial_transform(self, traj):
+        """Featurize an MD trajectory into a vector space via distance
+        after superposition
+
+        Parameters
+        ----------
+        traj : mdtraj.Trajectory
+            A molecular dynamics trajectory to featurize.
+
+        Returns
+        -------
+        features : np.ndarray, dtype=float, shape=(n_samples, n_features)
+            A featurized trajectory is a 2D array of shape
+            `(length_of_trajectory x n_features)` where each `features[i]`
+            vector is computed by applying the featurization function
+            to the `i`th snapshot of the input trajectory.
+
+        See Also
+        --------
+        transform : simultaneously featurize a collection of MD trajectories
+        """
+        traj.superpose(self.reference_traj, atom_indices=self.superpose_atom_indices)
+        x = []
+        for i in range (0, self.reference_traj.n_frames):
+            y = np.array([md.rmsd(traj, self.reference_traj, i)])
+            x = np.append(x,y)
+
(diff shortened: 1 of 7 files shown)


rmcgibbo:
Nice work. This looks like a great feature to add. I'll make a couple comments inline.

msmbuilder/featurizer/featurizer.py
```diff
@@ -200,6 +200,61 @@ def partial_transform(self, traj):
         return x
 
 
+
+class StrucRMSDFeaturizer(Featurizer):
+    """Featurizer based on overall RMSD to a set of reference structures.
+
```
rmcgibbo: Can you add the description that you included in the PR (about what this featurizer is computing) to the docstring. The first line is good, but you could add the rest as a more detailed description.

msmbuilder/featurizer/featurizer.py
```diff
@@ -200,6 +200,61 @@ def partial_transform(self, traj):
...
+    Parameters
+    ----------
+    atom_indices : np.ndarray, shape=(n_atoms,), dtype=int
+        The indices of the atoms to superpose and compute the distances with
+    reference_traj : md.Trajectory
+        The reference conformation to superpose each frame with respect to
+        (only the first frame in reference_traj is used)
+    superpose_atom_indices : np.ndarray, shape=(n_atoms,), dtype=int
```
rmcgibbo: This argument should probably be named `atom_indices` instead of `superpose_atom_indices`.

msmbuilder/featurizer/featurizer.py
```diff
@@ -200,6 +200,61 @@ def partial_transform(self, traj):
...
+
+        See Also
+        --------
+        transform : simultaneously featurize a collection of MD trajectories
+        """
+        traj.superpose(self.reference_traj, atom_indices=self.superpose_atom_indices)
+        x = []
+        for i in range(0, self.reference_traj.n_frames):
+            y = md.rmsd(traj, self.reference_traj, i)
+            x.append(y)
```
rmcgibbo: This `for` loop will be needlessly slow, because rmsd has been highly optimized but there's a lot of overhead for python looping. This is why numpy operations that act on entire vectors at once are much faster than pure python loops.


It's probably good enough for now, but we could do something faster. This calculation is analgous to a function called `cdist` in the scipy library ([docs](http://docs.scipy.org/doc/scipy-0.15.1/reference/generated/scipy.spatial.distance.cdist.html)). We have RMSD and Trajectory-aware analogs of many of the functions in that module in the subpackage `msmbuilder.libdistance` -- e.g. `pdist` [here](https://github.com/msmbuilder/msmbuilder/blob/master/msmbuilder/libdistance/libdistance.pyx#L132-L157), but it looks like we don't have one for cdist. 

------

For example, with this code on my laptop, 

```
# test_timing.ipy
import numpy as np

def add_arrays_python(a, b):
    assert len(a) == len(b)
    result = np.empty_like(a)
    size = len(a)

    for i in range(size):
        result[i] = a[i] + b[i]

    return result


def add_arrays_numpy(a, b):
    assert len(a) == len(b)

    result = a + b
    
a = np.random.randn(10000)
b = np.random.randn(10000)
%timeit add_arrays_python(a, b)
%timeit add_arrays_numpy(a, b)
```

```
$ ipython  test_timing.ipy
100 loops, best of 3: 3.57 ms per loop
The slowest run took 5.73 times longer than the fastest. This could mean that an intermediate result is being cached
100000 loops, best of 3: 6.84 µs per loop
```
So the numpy code is ~500x faster at computing the same result.

msmbuilder/featurizer/featurizer.py
brookehus: @rmcgibbo so what about the atom_indices (first parameter) above? I guess I could just rewrite all of this since I copied it over from superposefeaturizer but didn't really cross check. The description part has been updated though.

msmbuilder/featurizer/featurizer.py
brookehus: @rmcgibbo when I get back from thanksgiving can we look at how to do this the numpy way together, i.e. (1) incorporate cdist and (2) make things faster?

msmbuilder/featurizer/featurizer.py
rmcgibbo: Sorry -- I had this a little wrong. But the use of `superpose()` at line 251 is not having the effect that you think it is. `md.rmsd()` at line 254 computes the minimal rmsd over all rotations and translations -- it's invariant to any pre-rotation you might do.

msmbuilder/featurizer/featurizer.py
rmcgibbo: Absolutely.

brookehus: So we're looking for something like scipy.spatial.distance.cdist(traj, ref, 'rmsd'), and this would replace the entire loop because we'd never need to iterate by i. To do this we'd have to create the equivalent cdsit in msmbuilder where pdist is and also make the _cdist_rmsd, e.g.

Without doing this it will still work, but slowly, right? However, I'm not sure how to find out why one of the checks is failing.

msmbuilder/featurizer/featurizer.py
brookehus: Makes sense- I see that if I just comment out the superpose line nothing changes.

msmbuilder/featurizer/featurizer.py
```diff
@@ -200,6 +200,64 @@ def partial_transform(self, traj):
         return x
 
 
+
+class StrucRMSDFeaturizer(Featurizer):
+    """Featurizer which inputs a trajectory to be analyzed ('traj') and a
```
mpharrigan: First line of docstrings should be a one sentence summary

More details can go in a paragraph below

msmbuilder/libdistance/libdistance.pyx
```diff
@@ -126,6 +131,56 @@ def assign_nearest(X, Y, const char* metric, npy_intp[::1] X_indices=None):
...
+        The distance metric to use. metric = "rmsd" requires that X be of type
+        md.Trajectory; other distance metrics require that it be a numpy array
+
+    Returns
+    -------
+    Y : ndarray
+        A :math:`n_A` by :math:`n_B` distance matrix is returned.
+        For each :math:`i` and :math:`j`, the metric
+        ``dist(u=XA[i], v=XB[j])`` is computed and stored in the
+        :math:`ij` th entry.
```
rmcgibbo: I think this docstring on the return value hasn't need updated for the return value being two-dimensional

msmbuilder/libdistance/libdistance.pyx
```diff
@@ -126,6 +131,56 @@ def assign_nearest(X, Y, const char* metric, npy_intp[::1] X_indices=None):
...
+def cdist(XA, XB, const char* metric):
+    """cdist(XA, XB, metric):
+
+    Computes distance between each pair of the two collections of inputs.
+
+    Parameters
+    ----------
+    XA : array, shape = (n_samples, n_features) or md.Trajectory
+        The data array
+    XB : array, shape = (n_samples, n_features) or md.Trajectory
```
rmcgibbo: The `shape` documentation here should use a different variable for the length of XA and the length of XB, since they don't need to have the same length (but do need to have the same n_features).

msmbuilder/libdistance/libdistance.pyx
```diff
@@ -126,6 +131,56 @@ def assign_nearest(X, Y, const char* metric, npy_intp[::1] X_indices=None):
...
+    -------
+    Y : ndarray
+        A :math:`n_A` by :math:`n_B` distance matrix is returned.
+        For each :math:`i` and :math:`j`, the metric
+        ``dist(u=XA[i], v=XB[j])`` is computed and stored in the
+        :math:`ij` th entry.
+
+    Note
+    ----
+    Not implemented for X_indices != None
```
rmcgibbo: I'm not sure this note is meaningful to anyone else :)

rmcgibbo:
you might want to add to https://github.com/msmbuilder/msmbuilder/blob/master/msmbuilder/tests/test_libdistance.py too

msmbuilder/libdistance/libdistance.pyx
brookehus: @rmcgibbo if XA has na samples and XB has nb samples then isn't the matrix na * nb?

msmbuilder/featurizer/featurizer.py
```diff
@@ -200,6 +200,59 @@ def partial_transform(self, traj):
...
+            A featurized trajectory is a 2D array of shape
+            `(length_of_trajectory x n_features)` where each `features[i]`
+            vector is computed by applying the featurization function
+            to the `i`th snapshot of the input trajectory.
+
+        See Also
+        --------
+        transform : simultaneously featurize a collection of MD trajectories
+        """
+        result = cdist(traj, self.reference_traj, 'rmsd')
```
rmcgibbo: the `import` and module are missing here.

brookehus: @rmcgibbo what needs to be imported? :(

msmbuilder/libdistance/libdistance.pyx
```diff
@@ -356,6 +409,59 @@ cdef _assign_nearest_float(float[:, ::1] X, float[:, ::1] Y,
...
+    cdef float[::1] XA_trace, XB_trace
+    cdef double[:, ::1] out
+
+    if XA._rmsd_traces is None:
+        raise ValueError('XA must be pre-centered using '
+                         'md.Trajectory.center_coordinates')
+    if XB._rmsd_traces is None:
+        raise ValueError('XB must be pre-centered using '
+                         'md.Trajectory.center_coordinates')
+
```
brookehus: @rmcgibbo instead of this error message can I tell it to do XA.center_coordinates()? e.g.

+    if XA._rmsd_traces is None:
 +        XA.center_coordinates()

rmcgibbo: Sure. That's probably more user friendly.

rmcgibbo:
Sweet. This is great.

PR #696, opened 2015-11-30
[WIP] initial check in for ktica

msultan:
This is basically @evanfeinberg 's implementation of ktica using the nystorem landmark approximation for kernalizing the input data. I basically added a wrapper for the nystorem and another for the pipeline. Still need to add tests, docs, and the option to add custom landmarks.

### DIFF
--- a/msmbuilder/decomposition/ktica.py
+++ b/msmbuilder/decomposition/ktica.py
@@ -1,0 +1,203 @@
+# Author:  Evan Feinberg<enf@stanford.edu>
+# Contributors: Muneeb Sultan<msultan@stanford.edu>, Robert McGibbon <rmcgibbo@gmail.com>, \
+# Kyle A. Beauchamp  <kyleabeauchamp@gmail.com>
+# Copyright (c) 2014, Stanford University
+# All rights reserved.
+
+#-----------------------------------------------------------------------------
+# Imports
+#-----------------------------------------------------------------------------
+
+from __future__ import print_function, division, absolute_import
+import numpy as np
+import scipy.linalg
+import warnings
+from .tica import tICA
+from .kernal_approx import Nystroem
+from ..base import BaseEstimator
+from sklearn.pipeline import Pipeline
+from sklearn.base import TransformerMixin
+from ..utils import check_iter_of_sequences
+
+__all__ = [ 'ktICA']
+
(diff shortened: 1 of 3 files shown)


mpharrigan:
noice

rmcgibbo:
I'll let others make the call, but I'm -1. The file ``msmbuilder/decomposition/kernal_approx.py`` seems like a good addition -- +1 for that. But the other file doesn't add anything new. It's just a lot of complicated docstrings around a very simple `Pipeline`.

mpharrigan:
I'll take a closer look later

msmbuilder/decomposition/ktica.py
```diff
@@ -0,0 +1,203 @@
...
+
+        Returns
+        -------
+        sequence_new : list of array-like, each of shape (n_samples_i, n_components)
+
+        """
+        check_iter_of_sequences(sequences, max_iter=3)  # we might be lazy-loading
+
+        X_transformed = self._kt_pipeline.transform(sequences)
+        #n_x = self.nystorm.transform(X)
```
cxhernandez: Why are these here?

msultan: I am updating the docs. gimme a few hours, I was trying to add tests first.

msultan:
@all we could potentially just add a command line script and move the rest of the code to somewhere else. Any recommendations?

cxhernandez:
Yeah. Maybe make an example of this for the docs?

evanfeinberg:
Thanks, @msultan, for integrating this. It should be pretty straightforward for us to add custom landmark points. Please see this modified Nystroem class:

https://github.com/evanfeinberg/conformation/blob/master/LandmarkNystroem.py

So if we just import this instead of the standard Nystroem, the feature is available. If you don't pass the argument "basis", it generates a random set of basis points/landmarks, or you can pass a list of basis points.

mpharrigan:
@evanfeinberg have you considered proposing that change upstream to sklearn

msultan:
yea i agree with @mpharrigan's suggestion. Though if they dont want it, we can always have a local variant too.

mpharrigan:
Ok, I agree that we shouldn't do a weird pipeline thing to combine two steps

@msultan : do you want to re-work this pr and just include the nystroem wrapper?

PR #704, opened 2016-01-15
Rewrite SparseTICA using new algorithm

rmcgibbo:
- This PR adds a new solver for SparseTICA, which used to use cvxpy and didn't really work (it was always very experimental). The new solver uses ADMM and is much much faster and more robust. I have a manuscript in preparation describing this method, so it's probably not going to be clear from just looking at the code how everything works.
- It also changes the keyword arguments a little bit for `SparseTICA` and `tICA`. In particular, I've removed the `gamma` keyword argument and replaced it with a keyword argument called `shrinkage`. It does the same thing but has slightly different semantics -- I've also changed the default value to `None`. This does break the API.
- When the default, `None`, is used, the regularization strength is estimated analytically using the "Rao-Blackwellized Ledoit-Wolf" estimator from "Chen, Wiesel, and Hero III. ICASSP (2009)". I tried many of the analytic shrinkage covariance estimators from the literature (see library and experiments: https://pythonhosted.org/covar/). The RBLW estimator is the best, in my opinion, because it gives very good results and it is not any more expensive, unlike the LW estimator which is much more expensive to compute for tICA because it requires computing and storing some 4th moments, IIRC.


.gitignore
```diff
@@ -43,5 +43,5 @@ msmbuilder/version.py
 # Vim temp files
 *.swp
 *.swo
-
 build.log
+msmbuilder/decomposition/_speigh.cpp
```
mpharrigan: convention is to put this in `.gitignore` in `decomposition/` subdirectory

msmbuilder/decomposition/_speigh.pyx
```diff
@@ -0,0 +1,402 @@
...
+        # primal residual: r_norm2 = ||r||^2
+        cddot(r, r, &r_norm2)
+        cddot(s, s, &s_norm2)
+
+        if verbose > 1:
+            print(' rho', rho, ' residuals ', sqrt(r_norm2), sqrt(s_norm2))
+            #print('x', np.asarray(x))
+            #print('z', np.asarray(z))
+            #print('u', np.asarray(u))
+            #print('f', 0.5*np.dot(x,x) - np.dot(x,b) + 0.5*np.dot(b,b) + np.sum(np.abs(np.multiply(w, x))))
```
mpharrigan: maybe remove these?

rmcgibbo:
All code review is extremely welcome. I can add more comments and docstrings to describe what each function is doing, etc.

mpharrigan:
Did you forget to add the new gitignore?

rmcgibbo:
Yes.

msmbuilder/decomposition/sparsetica.py
```diff
@@ -123,49 +111,35 @@ class SparseTICA(tICA):
...
     """
 
-    def __init__(self, n_components, lag_time=1, gamma=0.05,
-                 rho=0.01, weighted_transform=True, epsilon=1e-6, tolerance=1e-8,
-                 maxiter=10000, max_nc=100, greedy=True, verbose=False):
-        super(SparseTICA, self).__init__(
-            n_components, lag_time=lag_time, gamma=gamma,
+    def __init__(self, n_components, lag_time=1, rho=0.01,
+                 weighted_transform=True, epsilon=1e-6, shrinkage=None,
+                 tolerance=1e-6, maxiter=10000, verbose=False):
```
brookehus: not having a default for n_components doesn't allow running SparseTICA() without arguments (which might be used e.g. in osprey). tICA can handle this and sets n_components=None by default and then later, if self.n_components is None: self.n_components = n_features

rmcgibbo: Done.

rmcgibbo:
Okay, I think the tests should pass on Linux this time around.

rmcgibbo:
@mpharrigan, @brookehus do you have any problems with me merging this now?

mpharrigan:
lgtm

PR #707, opened 2016-01-22
Refactor msm.transform into msm.partial_transform

mpharrigan:


mpharrigan:
wip: make sure this works with all the different subclasses of msm

mpharrigan:
Addresses #709 as well by deleting "trimmed_microstates_to_macrostates"

PR #708, opened 2016-01-22
Rebase parallelize libdistance

mpharrigan:
rebase of #354 

test_kcenters fails locally and test_kmedoids fails sometimes locally for me with these changes


mpharrigan:
Eh, something is legitimately wrong here (I think) and I can't figure it out

PR #711, opened 2016-01-23
 Implement bootstrap msm

msultan:
This is the initial commit for adding a msmb class for creating families of estimators via bootstraping over sequence space. 
I think this is what the api should look like. 

```python
from msmbuilder.msm import MarkovStateModel
from msmbuilder.msm.bootstrapmodel import BootstrapModel
msm = MarkovStateModel
bmsm = BootstrapModel(estimator = msm, n_samples = 10)
bmsm.fit(sequences)
```


msultan:
Ideally, this class would contain methods for comparing the mle estimator to the bootstrapped models and a nice mapping scheme from input space to all model space.

cxhernandez:
Just organizationally, I feel like this should be it's own submodule under `msmbuilder`, e.g.:

`from msmbuilder.bootstrapmodel import BootstrapModel `

or

`from msmbuilder import BootstrapModel `

msultan:
alright, i have changed it a bit so that the ```BootStrapMarkovStateModel``` mdl fits both the mle MSM and the bootstraped mdls in one go and adds some convenience attributes for comparing populations across all mdls. 
```
python
from msmbuilder.msm.validation import BootStrapMarkovStateModel
%pylab inline
bsmsm = BootStrapMarkovStateModel(lag_time=10,verbose=False,n_procs=2)
bsmsm.fit(seq)
```

I also added some 
errorbar(range(bsmsm.mle.n_states_),bsmsm.mapped_populations_mean_, 
         bsmsm.mapped_populations_sem_)

msultan:
okay i think this is ready for code review.

mpharrigan:
remove wip :)

msmbuilder/msm/validation/bootstrapmsm.py
```diff
@@ -0,0 +1,135 @@
...
+    def __init__(self, n_samples=10,  n_procs=None, **kwargs):
+        self.n_samples = n_samples
+        self.n_procs = n_procs
+        self.mle = MarkovStateModel(**kwargs)
+
+        self.all_populations_ = None
+        self.mapped_populations_ = None
+        self.resample_ind_ = None
+
+    def fit(self, sequences, y=None):
```
cxhernandez: Maybe there should be a `parallel` flag, which is default `True`?

msultan: Wouldn't n_procs=1 serve the same purpose? I am tempted to default to multiprocessing so that this fits in a somewhat reasonable time for larger datasets.

cxhernandez: > Wouldn't n_procs=1 serve the same purpose? 

Ah, that's true. Nevermind then.

> I am tempted to default to multiprocessing

Go for it :smile:

msultan:
fantastic, py2.7 cant pickle the _fit_one method. Gonna try and fix this later in the day.

msmbuilder/msm/validation/__init__.py
```diff
@@ -0,0 +1,2 @@
+from __future__ import absolute_import
+from .bootstrapmsm import *
```
mpharrigan: Don't import `*`
(this pulls in Pool, cpu_count, list_of_1d, MarkovStateModel et al. as well)
(you could define `__all__` in bootstrapmsm.py but why not just import the one class explicitly)

msmbuilder/msm/validation/bootstrapmsm.py
```diff
@@ -0,0 +1,164 @@
...
+
+    def _fit_one(self, sequences):
+        mdl = MarkovStateModel(**self.msm_args)
+        #there is no guarantee that the mdl fits this sequence set so
+        #we return None in that instance.
+        try:
+            mdl.fit(sequences)
+            # solve the eigensystem
+            mdl.eigenvalues_[0]
+        except:
```
mpharrigan: No bare `try...except`. This might hide real issues! Be specific about what exception you're expecting. If it makes sense, add a warning as well

msmbuilder/tests/test_bootstrap_msm.py
```diff
@@ -0,0 +1,90 @@
...
+    mle_mdl = MarkovStateModel(lag_time=1)
+    b_mdl = BootStrapMarkovStateModel(n_samples=10, n_procs=2, msm_args={'lag_time': 1})
+    mle_mdl.fit(seq)
+    b_mdl.fit(seq)
+    #make sure we have good model
+    eq(mle_mdl.populations_, b_mdl.mle_.populations_)
+    eq(mle_mdl.timescales_, b_mdl.mle_.timescales_)
+
+
+class fakemsm():
```
mpharrigan: inherit from `object` for py2

msmbuilder/tests/test_bootstrap_msm.py
```diff
@@ -0,0 +1,90 @@
...
+
+    pop = {0:0.1, 1:0.2, 2:0.3}
+    mdl1 = fakemsm(3, mdl1_mapping, pop)
+    mdl2 = fakemsm(2, mdl2_mapping, pop)
+
+    mapped_pop = mapper(mdl1, mdl2)
+    assert (mapped_pop==[0.1, 0, 0.3]).all()
+
+def test_mapper_4():
+   #case where the mdl is jumbled around
```
mpharrigan: indent this to the same level

msmbuilder/msm/validation/bootstrapmsm.py
msultan: ohh good point @mpharrigan . i was thinking of the value error that pops up when fitting msms every now and then. Gonna change it to explicitly expect the value error. .

msmbuilder/msm/validation/bootstrapmsm.py
```diff
@@ -0,0 +1,168 @@
...
+            mdl2_mapped = mdl2.mapping_[mdl1_unmapped]
+            return_vect[i] =  mdl2.populations_[mdl2_mapped]
+        except:
+            pass
+    return return_vect
+
+
+
+def _fit_one(jt):
+    print(jt)
```
mpharrigan: is there a reason why you're printing here?

msultan: shit, forgot about that. Was debugging on py27. Thanks for the catch

mpharrigan:
:+1:

msultan:
Merged!

Issue #706, opened 2016-01-22
runtests.py doesn't work


mpharrigan:
```
Building, see build.log...
Build OK
Traceback (most recent call last):
  File "runtests.py", line 251, in <module>
    main(argv=sys.argv[1:])
  File "runtests.py", line 123, in main
    test = sys.modules[PROJECT_MODULE].test
AttributeError: module 'msmbuilder' has no attribute 'test'
```

changing that line to `tests` (which should exist) doesn't work (gives same error)

PR #714, opened 2016-01-28
Re-do runtests.py

mpharrigan:
addresses #706

### DIFF
--- a/msmbuilder/tests/test_msm.py
+++ b/msmbuilder/tests/test_msm.py
@@ -357,7 +357,7 @@ def test_score_1():
         assert_approx_equal(model.score([sequence]), model.eigenvalues_.sum())
         assert_approx_equal(model.score([sequence]), model.score_)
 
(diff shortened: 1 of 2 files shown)


mpharrigan:
This simplifies test discovery/running by just using `subprocess` with a modified `PYTHONPATH`

rmcgibbo:
why not just add an `msmbuilder.test()` function like mdtraj?

mpharrigan:
I couldn't get that to work (tried it). 

I suspect the fact that `import msmbuilder` doesn't import everything makes the test discovery not work

rmcgibbo:
> I suspect the fact that import msmbuilder doesn't import everything makes the test discovery not work
 Hide all checks

I doubt that, based on the existence of `scipy.test()`. But it's not important.

mpharrigan:
Ok, I was probably doing something wrong, but I couldn't get programmatic test discovery to work :)

This does work though!

PR #717, opened 2016-01-29
Function featurizer

msultan:
Featurizer capable of hopefully taking any arbitrary function and mapping it to the input set of trajectories. I wrote one test though i dont know how this is gonna get put into the command line.


mpharrigan:
Let's make the api *only* accept a function (/callable).

If we want to expose this on the command line, then we can include a thing in the command-line parser to load from a pickle file

msmbuilder/featurizer/featurizer.py
```diff
@@ -307,6 +309,56 @@ def partial_transform(self, traj):
...
+    ----------
+    function : instatiation of the function
+
+    """
+
+    def __init__(self, function, **kwargs):
+        if hasattr(function, '__call__'):
+            self.function = function
+            self.args = {}
+            self.args.update(kwargs)
```
mpharrigan: why not `self.args = kwargs`?

msmbuilder/featurizer/featurizer.py
```diff
@@ -307,6 +309,56 @@ def partial_transform(self, traj):
...
+        elif os.path.isfile(function):
+            self.function = verboseload(function)
+        else:
+            raise Exception("Sorry but we "
+                            "couldn't use the provided"
```
mpharrigan: Missing a space

msmbuilder/featurizer/featurizer.py
```diff
@@ -307,6 +309,56 @@ def partial_transform(self, traj):
...
+                            "function.")
+
+    def partial_transform(self, traj):
+        """Featurize an MD trajectory into a vector space via
+        the input function
```
mpharrigan: First line of docstring should be one complete sentence (aka, don't wrap here)

msmbuilder/tests/test_featurizer.py
```diff
@@ -48,6 +50,30 @@ def test_SubsetAtomPairs2():
...
+    trj0 = trajectories[0][0]
+    #use the dihedral to compute phi for ala
+
+
+    atom_ind = [[4, 6, 8,14]]
+    func = compute_dihedrals
+    f = FunctionFeaturizer(func, indices=atom_ind)
+    res1 = f.transform([trj0])
+
+    def funcception(trj):
```
mpharrigan: lol

msmbuilder/tests/test_featurizer.py
```diff
@@ -48,6 +50,30 @@ def test_SubsetAtomPairs2():
...
+        return compute_phi(trj)[1]
+
+    f = FunctionFeaturizer(funcception)
+    res2 = f.transform([trj0])
+
+    f3 = DihedralFeaturizer(['phi'], sincos=False)
+    res3 = f3.transform([trj0])
+
+    assert res1==res2==res3
```
mpharrigan: Does this work? Shouldn't you use `np.testing.assert_array_equal` or something like that to check elementwise equality

msmbuilder/tests/test_featurizer.py
```diff
@@ -48,6 +50,30 @@ def test_SubsetAtomPairs2():
...
+    return
```
mpharrigan: no return

msmbuilder/featurizer/featurizer.py
```diff
@@ -307,6 +307,73 @@ def partial_transform(self, traj):
...
+            A featurized trajectory is a 2D array of shape
+            `(length_of_trajectory x n_features)` where each `features[i]`
+            vector is computed by applying the featurization function
+            to the `i`th snapshot of the input trajectory.
+
+        See Also
+        --------
+        transform : simultaneously featurize a collection of MD trajectories
+
+         Notes:
```
mpharrigan: - remove space before this
- no colon
- remove extra underbar

msmbuilder/featurizer/featurizer.py
```diff
@@ -307,6 +307,73 @@ def partial_transform(self, traj):
...
+class FunctionFeaturizer(Featurizer):
+    """Featurizer based on arbitrary functions.
+
+    This featurizer transforms a dataset containing MD trajectories into
+    a vector dataset by representing each frame in each of the MD trajectories
+    by a vector the output of the function.
+
+    Parameters
+    ----------
+    function : Instantiation of the function
```
mpharrigan: use the numpydoc format
```
name : type
    description
```

msmbuilder/featurizer/featurizer.py
```diff
@@ -307,6 +307,73 @@ def partial_transform(self, traj):
...
+    *args : arguments to pass to the function MINUS the trajectory keyword
+    **kwargs: key word arguments to pass to the function MINUS the trajectory
+    keyword
+
+    UseCase:
```
mpharrigan: This is not a valid section

msultan: Should i move it to the notes? I feel having an example here is going to be useful.

mpharrigan: I think there's something like "example". check https://github.com/numpy/numpy/blob/master/doc/HOWTO_DOCUMENT.rst.txt

msmbuilder/featurizer/featurizer.py
```diff
@@ -307,6 +307,75 @@ def partial_transform(self, traj):
...
+    Parameters
+    ----------
+    function : function
+        Instantiation of the function. The function should accept
+        a mdtraj.Trajectory object as the first argument.
+    *args : list
+        arguments to pass to the function MINUS the trajectory keyword
+    **kwargs: kwargs
+        key word arguments to pass to the function MINUS the trajectory
+    keyword
```
mpharrigan: indent

msmbuilder/featurizer/featurizer.py
```diff
@@ -307,6 +307,77 @@ def partial_transform(self, traj):
...
+
+    Examples
+    --------
+    >>> function = compute_dihedrals
+    >>> f = FunctionFeaturizer(function, indices=[[0,1,2,3]])
+    >>> results = f.transform(dataset)
+    """
+
+    def __init__(self, function, *args, **kwargs):
+        if hasattr(function, '__call__'):
```
mpharrigan: I think using the builtin callable covers more use-cases here

https://docs.python.org/3.4/library/functions.html#callable

mpharrigan:
For the api, I think it would be more robust for the `__init__` method to take a dictionary of keyword arguments rather than `**kwargs`

```python
FunctionFeaturizer(func=mdtraj.compute_dihedrals, func_args={'indices': [[0,1,2,3]]})
```

I can see the current scheme messing up (1) the command line generation and (2) sklearn pipelines, both of which rely on introspection and don't expect `*args` or `**kwargs`

msmbuilder/featurizer/featurizer.py
msultan: ohh, didnt know about that.

msultan:
> 
```python
FunctionFeaturizer(func=mdtraj.compute_dihedrals, func_args={'indices': [[0,1,2,3]]})
```
alright, i will switch it.

msultan:
hmm weird, tests seem to be failing for at random places.

msmbuilder/featurizer/featurizer.py
```diff
@@ -307,6 +307,75 @@ def partial_transform(self, traj):
...
+        if callable(function):
+            self.function = function
+            self.func_args = func_args
+        else:
+            raise Exception("Sorry but we "
+                            "couldn't use the provided "
+                            "function.")
+
+    def partial_transform(self, traj):
+        """Featurize a MD trajectory into a vector by
```
mpharrigan: Can you make this a one-line summary?

msmbuilder/featurizer/featurizer.py
```diff
@@ -307,6 +307,75 @@ def partial_transform(self, traj):
...
+
+        Notes
+        -----
+        This method assumes that the function takes in the trajectory object
+        as the first argument.
+
+        """
+        x = []
+        x.append(self.function(traj,  **self.func_args))
+        return np.hstack(x)
```
mpharrigan: These three lines are redundant. Just return `self.function(traj, **self.func_args)`

msmbuilder/featurizer/featurizer.py
```diff
@@ -307,6 +307,75 @@ def partial_transform(self, traj):
...
+    >>> f = FunctionFeaturizer(function, func_args={indices: [[0,1,2,3]]})
+    >>> results = f.transform(dataset)
+    """
+
+    def __init__(self, function, func_args={}):
+        if callable(function):
+            self.function = function
+            self.func_args = func_args
+        else:
+            raise Exception("Sorry but we "
```
mpharrigan: This should probably be `ValueError` and maybe include the fact that the object isn't callable in the error message so the user knows what's wrong

cxhernandez: +1

msultan: that makes more sense. Making the changes now.

mpharrigan:
:+1:

msultan:
aww i wanted to merge this

PR #721, opened 2016-02-02
Be aggressive against argspec deprecation warnings

mpharrigan:


rmcgibbo:
This is actually hilarious.

msultan:
maybe email the warnings as text to argspec's owners.

rmcgibbo:
Well, once Anaconda updates their packages to the latest releases, it'll stop being a problem. It's already been fixed upstream.

mpharrigan:
It would also help if sklearn didn't force warnings to "on" with no way to turn it off

rmcgibbo:
Where in the sklearn code is that done?

mpharrigan:
Ok, I don't think sklearn does it anymore but something in msmbuilder/mdtraj is turning warnings on!

```python
>>> import warnings
>>> warnings.warn("hi", DeprecationWarning)
>>> import mdtraj
>>> warnings.warn("hi", DeprecationWarning)
__main__:1: DeprecationWarning: hi
```

Run the above from not-ipython (which also fiddles with default warnings)

mpharrigan:
@rmcgibbo !! for shame

https://github.com/mdtraj/mdtraj/blob/master/mdtraj/utils/__init__.py#L22

mpharrigan:
This shouldn't be necessary with new mdtraj release

Issue #539, opened 2015-04-08
Indexing semantics for dataset are all wrong


rmcgibbo:
Is it like a list with "keys" as contiguous indices and a `len`? Or are they filenames, integers? It's kind of a mess right now.

mpharrigan:
+1 for figuring this out

kyleabeauchamp:
I guess I'd vote for something that feels like pandas

mpharrigan:
I think it should be pandas-y too, but we may have to convince @rmcgibbo :panda_face:

mpharrigan:
@msultan is writing a "pull structures" script right now where he keeps track of trajectories in a dictionary keyed on trajectory filename. This involves patterns like:
```python
transformed_dict = {}
for k in traj_dict:
   transformed_dict[k] = tica.partial_transform(traj_dict[k])
```

mpharrigan:
I've sketched out my ideal view of dealing with datasets in this gist
https://gist.github.com/mpharrigan/c2b68d8d015133f1e387

mpharrigan:
I'm working incrementally towards improving this with #610 and #722 

Datasets should be dictionary-like with meaningful keys

cxhernandez:
@mpharrigan: was this done in #722 and #768?

PR #722, opened 2016-02-04
Use tuples of ints as keys to dataset objects

mpharrigan:
Right now only the dir-npy dataset, lets you do stuff like this:

```python
ds[9752, 5, 1] = np.array()
```

```bash
ls datasetname/09752/00005/00001.npy
```

Which is nicer for organizing fah-style data.

It should only be used with integers [or things formattable by `"{:d}"` and sortable]

I'm thinking of adding a flag to disable 0-padding of the filenames. The sorting should be done with the regular-expressioned integer values anyways.

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -15,6 +15,10 @@ API Changes
 New Features
 ~~~~~~~~~~~~
 
+- The NumpyDirDataset can accept tuples of integers as keys, and will create
+  a directory structure as appropriate. `ds[9752, 0, 1] = np.array()` will
+  put a file called `000000001.npy` under the directory `dsname/09752/00000/`
+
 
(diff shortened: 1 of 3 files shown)


mpharrigan:
cc #539

mpharrigan:
Added a kwarg `zero_pad`, which can be set to false to use filenames without all the zeros

mpharrigan:
I think this isn't quite the right idea. See #768 for a better sense of where I want to take this.

Specifically, I don't like having to parse arbitrary keys from filenames. I'd much rather have a persistent list of keys (e.g. as the index of a dataframe) that are used to generate filenames.

key ----> filename :white_check_mark: 
filename ----> key :warning: :skull:

PR #724, opened 2016-02-05
Bootstrap with scores

brookehus:
Utilizes the fact that bootstrapping already splits the data into a sampled and omitted set to train an msm on the sampled set of trajectories and simultaneously test it on the omitted set. This PR provides a score for the training and test set and means and std's of those scores.

### DIFF
--- a/msmbuilder/msm/validation/bootstrapmsm.py
+++ b/msmbuilder/msm/validation/bootstrapmsm.py
@@ -1,5 +1,5 @@
 # Author: Mohammad M. Sultan <msultan@stanford.edu>
-# Contributors:
+# Contributors: Brooke Husic <brookehusic@gmail.com>
 # Copyright (c) 2016, Stanford University
 # All rights reserved.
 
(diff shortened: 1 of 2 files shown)


msmbuilder/msm/validation/bootstrapmsm.py
```diff
@@ -67,6 +82,11 @@ class BootStrapMarkovStateModel(_MappingTransformMixin):
     of workers making it capable of parallelizing across
     compute nodes via mpi or ipyparallel. This can lead to
     a significant speed up for larger number of samples.
+
+    Example
```
mpharrigan: i think this is supposed to be example*s*

mpharrigan: I think this is the canonical list of allowed sections

https://github.com/numpy/numpydoc/blob/master/numpydoc/docscrape.py#L105

mpharrigan:
Do you want to add a small test for this?

brookehus:
@mpharrigan boo. it passed before i added the test.

mpharrigan:
I restarted it. We should make it so these tests don't randomly fail

PR #725, opened 2016-02-06
removed print statement from bootstrap msm and added…

msultan:
1. Removed a print statement for the bootstrapping code. 

2. Added a hidden attribute that keeps track of ommitted trajectories. 

3. I also added a test for making sure the resampling is working though that test has 1 × 10^-200 probability of randomly failing at some point.


mpharrigan:
Why did you choose to hide the new attribute?

msultan:
Didnt seem like it was really necessary to expose it.

Issue #718, opened 2016-01-29
Move counts setup into a new mixin


mpharrigan:
https://github.com/msmbuilder/msmbuilder/blob/master/msmbuilder/msm/core.py#L29

is not related to the functionality of a mapping transformer; it should be a new mixin

rmcgibbo:
I think the rate matrix code might be out of sync on how it handles ergodic triming, so this could help with that too

mpharrigan:
I think @msultan may have addressed that (?) when he refactored everything into the mapping transformer; it's just not the right name imo

rmcgibbo:
No I don't think so.

msultan:
I only changed the msm and bayesian msm modules. The others are still using their respective counts code.

mpharrigan:
Good to know. Then yes, let's move everything to the same counts implementation

msultan:
closing this for now.

PR #729, opened 2016-02-12
same counts setup

msultan:
Based upon #718, I refactored the build counts setup into a new mixin. Naming could be better but I wasn't sure what else to call it. Also added a test that compares the counts across 3 models and gets the same answer. Might need to construct a test that throws some states away but this should suffice for now i think.

### DIFF
--- a/msmbuilder/msm/core.py
+++ b/msmbuilder/msm/core.py
@@ -25,30 +25,6 @@ __all__ = [
 
 
 class _MappingTransformMixin(TransformerMixin):
-    def _setup(self, sequences, y=None):
-        sequences = list_of_1d(sequences)
-        # step 1. count the number of transitions
-        if int(self.lag_time) <= 0:
-            raise ValueError('Invalid lag_time: %s' % self.lag_time)
-        raw_counts, mapping = _transition_counts(
-                sequences, int(self.lag_time),
-                sliding_window=self.sliding_window)
-
-        ergodic_cutoff = self._parse_ergodic_cutoff()
-        if ergodic_cutoff > 0:
-            # step 2. restrict the counts to the maximal strongly ergodic
-            # subgraph
-            self.countsmat_, mapping2, self.percent_retained_ = \
-                _strongly_connected_subgraph(raw_counts, ergodic_cutoff,
-                                             self.verbose)
-            self.mapping_ = _dict_compose(mapping, mapping2)
-        else:
-            # no ergodic trimming. 
-            self.countsmat_ = raw_counts
-            self.mapping_ = mapping
-            self.percent_retained_ = 100
-        self.n_states_ = self.countsmat_.shape[0]
-        return ergodic_cutoff
 
     def partial_transform(self, sequence, mode='clip'):
         """Transform a sequence to internal indexing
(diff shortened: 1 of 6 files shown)


mpharrigan:
Very nice!

msmbuilder/msm/ratematrix.py
```diff
@@ -131,30 +132,6 @@ def __init__(self, lag_time=1, n_timescales=None, ergodic_cutoff=1,
         self.right_eigenvectors_ = None
         self.percent_retained_ = None
```
mpharrigan: don't you need to add in a call to _build_counts in continuoustimemsm

mpharrigan: nevermind, i saw it

msmbuilder/msm/core.py
```diff
@@ -205,6 +181,32 @@ def inverse_transform(self, sequences):
...
+                _strongly_connected_subgraph(raw_counts, ergodic_cutoff,
+                                             self.verbose)
+            self.mapping_ = _dict_compose(mapping, mapping2)
+        else:
+            # no ergodic trimming.
+            self.countsmat_ = raw_counts
+            self.mapping_ = mapping
+            self.percent_retained_ = 100
+        self.n_states_ = self.countsmat_.shape[0]
+        return ergodic_cutoff
```
mpharrigan: why does this return `ergodic_cutoff`? Maybe just don't return anything

msmbuilder/msm/msm.py
```diff
@@ -154,7 +156,7 @@ def fit(self, sequences, y=None):
         None will not be counted. The mapping_ attribute will not include the
         NaN or None.
         """
-        _ = self._setup(sequences)
+        _ = self._build_counts(sequences)
```
mpharrigan: might as well completely omit the `_ = `

(the reason why this is sometimes used is if you are unpacking a tuple and don't care about one of the values)
```python
five, _ = (5, SomeCrazyObject())
print(five)
```

msmbuilder/msm/core.py
msultan: Thats true, let me see if the tests all pass and i can push another change in.

msmbuilder/msm/msm.py
msultan: I think this would change once i update the ```_build_counts ``` method.

mpharrigan:
And you can add an appropriate entry to `changelog.rst` explaining how counts used to be inconsistent and now they will always be consistent

msultan:
also yay/nay on the naming scheme?

mpharrigan:
Seems fine to me (aka yay)

msultan:
maybe we should just call it MSMMixin and move the sample_discrete method there too?

msultan:
why the hell is ```test_kmedoids.py``` failing on py3.5?

mpharrigan:
> maybe we should just call it MSMMixin and move the sample_discrete method there too?
-1 for now

msultan:
@mpharrigan any idea as to why this is failing on py3.4 but passing on py2.7 and py3.5?

mpharrigan:
If you create a py34 conda env can you reproduce locally?

I'll restart the tests but it may be a real issue

mpharrigan:
eh probably just something weird

PR #736, opened 2016-02-16
StrucRMSDFeaturizer now respects atom indices

brookehus:
Now it uses all atoms if no `atom_indices` are provided, and slices the trajectory appropriately otherwise. There are no `superpose_atom_indices` anymore; the featurizer requires them to be the same as the regular `atom_indices`. Also updated the relevant test.

### DIFF
--- a/msmbuilder/tests/test_strucrmsdfeaturizer.py
+++ b/msmbuilder/tests/test_strucrmsdfeaturizer.py
@@ -1,45 +1,65 @@
 import numpy as np
-
+import mdtraj as md
 from msmbuilder.example_datasets import fetch_alanine_dipeptide
 from msmbuilder.featurizer import StrucRMSDFeaturizer
 
+# np.testing.assert_array_almost_equal(array1,array2)
 
-def test_alanine_dipeptide():
+def test_alanine_dipeptide_basic():
     # This test takes the rmsd of the 0th set of alanine dipeptide
     # trajectories relative to the 0th frame of the dataset.
-    # The test asserts that the first rmsd calculated will be zero
-    # for both all atom indices and a subset of atom indices.
-    # It then verifies that the RMSD's calculated from the different
-    # sets of atom indices are not the same.
+    # The test asserts that all rmsd's calculated will be equal
+    # to the ones that would be calculated straight from mdtraj.
 
     dataset = fetch_alanine_dipeptide()
     trajectories = dataset["trajectories"]
-    featurizer = StrucRMSDFeaturizer(trajectories[0][0],
-                                     np.arange(trajectories[0].n_atoms))
-    data = featurizer.transform(trajectories[0])
+    featurizer = StrucRMSDFeaturizer(trajectories[0][0])
+    data = featurizer.transform(trajectories[0:1])
 
-    assert (data[0] < 1e-3)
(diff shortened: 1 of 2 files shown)


msmbuilder/tests/test_strucrmsdfeaturizer.py
```diff
@@ -7,18 +7,28 @@
...
                                      np.arange(trajectories[0].n_atoms))
     data = featurizer.transform(trajectories[0])
 
     assert (data[0] < 1e-3)
 
+    featurizer = StrucRMSDFeaturizer(trajectories[0][0],
+                                     range(trajectories[0].n_atoms)[0:11])
+    data_sliced = featurizer.transform(trajectories[0])
+
+    assert (data_sliced[0] < 1e-3)
```
mpharrigan: assert is a statement, not a function so typically the style is not to include outer parentheses

msmbuilder/tests/test_strucrmsdfeaturizer.py
```diff
@@ -38,10 +48,12 @@ def test_two_refs():
 
     dataset = fetch_alanine_dipeptide()
     trajectories = dataset["trajectories"]
-    featurizer = StrucRMSDFeaturizer(trajectories[0], trajectories[0][0:2],
+    featurizer = StrucRMSDFeaturizer(trajectories[0][0:2],
                                      range(trajectories[0].n_atoms))
     data = featurizer.transform(trajectories[0])
 
     # TODO: Figure out why arrays are 3D
```
mpharrigan: Can you address this todo?

mpharrigan: (I don't really know what it means). If it's no longer relevant you can get rid of it

msmbuilder/tests/test_strucrmsdfeaturizer.py
```diff
@@ -7,18 +7,28 @@
...
     trajectories = dataset["trajectories"]
-    featurizer = StrucRMSDFeaturizer(trajectories[0], trajectories[0][0],
+    featurizer = StrucRMSDFeaturizer(trajectories[0][0],
                                      np.arange(trajectories[0].n_atoms))
     data = featurizer.transform(trajectories[0])
 
     assert (data[0] < 1e-3)
 
+    featurizer = StrucRMSDFeaturizer(trajectories[0][0],
+                                     range(trajectories[0].n_atoms)[0:11])
```
mpharrigan: look up the signature for `np.arange` and consider using the start, stop arguments to construct the input you want rather than slicing a range

msmbuilder/tests/test_strucrmsdfeaturizer.py
```diff
@@ -7,18 +7,28 @@
...
+    data_sliced = featurizer.transform(trajectories[0])
+
+    assert (data_sliced[0] < 1e-3)
+    assert (data[1] is not data_sliced[1])
```
mpharrigan: http://stackoverflow.com/questions/132988/is-there-a-difference-between-and-is-in-python

mpharrigan:
It would be nice to include a test that verifies omitting the atom_indices argument works as expected

mpharrigan:
Fixes #734

PR #737, opened 2016-02-17
Test to make sure things show up on commandline

mpharrigan:
This set of autogenerated tests looks for things that get imported
in the msmbuilder subpackages (`msmbuilder.featurizer` , `.msm`,
`.decomposition`, etc) and tries to figure out whether there are classes
that probably should be exposed on the command line.

The tests test whether they're actually exposed on the command line.

The 'auto discovery' process isn't flawless, and you can manually
include or exclude classes from consideration. If there is a lot of
garbage being imported into these top level namespaces, I think we
really should consider fixing the root of the problem (don't import `*`)
rather than going crazy with the "exclude" section.

Issue #730, opened 2016-02-12
Get rid of references to mixtape


cxhernandez:
I first stumbled upon this in the [PCCA docstrings](https://github.com/msmbuilder/msmbuilder/blob/0183f0076535b3d34ba5038a2a528dc6f091ba6c/msmbuilder/lumping/pcca.py#L106): there are still quite a few [references to mixtape](https://github.com/msmbuilder/msmbuilder/search?utf8=%E2%9C%93&q=mixtape).

mpharrigan:
This is still unfixed

cxhernandez:
Oh woops. I though this was part of your sundry changes.

PR #738, opened 2016-02-17
Sundry changes to featurization

mpharrigan:
- move `subset_featurizer` to `subset` and don't import it into top-level `msmbuilder.featurizer` (somewhat cc #737 ) except for `get_atompair_indices`, which seems mildly useful. 
- remove `FirstSlicer`, fold functionality into `Slicer`
- remove `save` and `load` from featurizers
- fix reference to mixtape #730 
- fix runtests.py to prepend the test environment's bindir to PATH (for tests that use the command line)

These are breaking changes that probably should be done with a deprecation cycle, but I'm fairly certain no one is using this functionality


mpharrigan:
@msmbuilder/developers do any of you have any comments on these changes

rmcgibbo:
+1

msultan:
Seems fine to me.

Issue #741, opened 2016-02-23
von Mises option for DihedralFeaturizer


cxhernandez:
@rmcgibbo had a great idea to "soft bin" torsion angle data using multiple von Mises distributions with means equally spaced-out around a circle. This might be nice to add to MSMBuilder as an alternative to processing with sin/cos.

PR #743, opened 2016-02-25
Added von Mises option to DihedralFeaturizer

cxhernandez:
re: https://github.com/msmbuilder/msmbuilder/issues/741

adapted from https://gist.github.com/rmcgibbo/b4d218d229b32b7ace8a

### DIFF
--- a/msmbuilder/featurizer/featurizer.py
+++ b/msmbuilder/featurizer/featurizer.py
@@ -12,6 +12,7 @@ import mdtraj as md
 from sklearn.base import TransformerMixin
 import sklearn.pipeline
 from sklearn.externals.joblib import Parallel, delayed
+from scipy.stats import vonmises as vm
 from msmbuilder import libdistance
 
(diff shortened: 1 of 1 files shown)


cxhernandez:
Not sure if this is the optimal way to set up `DihedralFeaturizer`.... Right now, it requires you to explicitly assign `sincos` to `False` to use the `vonmises` option. Any thoughts?

mpharrigan:
Why not add a new featurizer?

Please put these branches on your fork so we can keep msmbuilder/msmbuilder tidy

Some of those auto-pep8 linebreaks are really jarring and pollute the diff

cxhernandez:
>Why not add a new featurizer?

I figured since it's a dihedral feature, it made sense to put it in the dihedral featurizer. But I'll make a new PR as it's own thing.

>Some of those auto-pep8 linebreaks are really jarring and pollute the diff

I'll try to tidy them up a bit.

PR #747, opened 2016-03-05
Sampling along tica coordinates

msultan:
 - [x] Add tests
 - [ ] Update changelog

New module to sample and interpret decomposition results via sampling from the trajectories. Added a test to check output type but need to add test to make sure it is sampling the right region .


msultan:
figshare is down and causing issues.

msultan:
still need to add

- [ ] String sampler to connect arbitrary regions 
- [ ] Coloring method for vmd

mpharrigan:
This needs a better name

msultan:
almost definitely, my idea was ```sample_dimension```, ```sample_region``` and ```connect_regions``` but all 3 are too generic. We should brainstorm this a bit more.

cxhernandez:
`sample_from_decomposition` or `sample_along_component`?

msultan:
@mpharrigan my understanding is that this is now implemented via #768 ?

mpharrigan:
Correct

PR #754, opened 2016-03-11
fix VonMisesFeaturizer

cxhernandez:
Not sure how it passed the tests before, but the ndarray shapes were set up wrong.


mpharrigan:
Can you add a test that would have failed the other one

mpharrigan:
bump!

mpharrigan:
Oh this is done!

Issue #752, opened 2016-03-09
implement MAP estimate for the bayesian models. 


msultan:
We need to be able to get the MAP estimate from the set of models being sampled from posteriors by all the Bayesian models to make the analysis somewhat tractable. i think the api should be something like 

``` python
mdl = Bayesianmodel()
mdl.fit(data)
mdl.map_estimate_ 
```

msultan:
Steps to do this
- [ ] Get the liklihood "energy"
- [ ] Use the maximum of the liklihood energy as the map estimate 
- [ ] the map estimate is the mle estimate conditioned on a uniform prior over the parameters. I am not sure if this is true in either of the Bayesian models.

PR #755, opened 2016-03-12
MAP estimator for bayes

msultan:
 - [ ] Update changelog
 - [ ] Test added

MAP estimator based upon issue raised in #752 . I made the choice of using the mean of the posterior logliklihoods though an argument could be made for the median of the distribution too based upon 
https://en.wikipedia.org/wiki/Maximum_a_posteriori_estimation

### DIFF
--- a/msmbuilder/msm/bayesmsm.py
+++ b/msmbuilder/msm/bayesmsm.py
@@ -45,7 +45,7 @@ from .core import (_MappingTransformMixin,
                    _CountsMSMMixin,
                    _solve_msm_eigensystem)
 from ._metzner_mcmc_fast import metzner_mcmc_fast
-from ._metzner_mcmc_slow import metzner_mcmc_slow
+from ._metzner_mcmc_slow import metzner_mcmc_slow, _logprob_T
 
 
 #-----------------------------------------------------------------------------
@@ -127,6 +127,12 @@ class BayesianMarkovStateModel(BaseEstimator, _MappingTransformMixin,
         matrix.
     transmats_ : array_like, shape = (n_samples, n_states_, n_states_)
         Samples from the posterior ensemble of transition matrices.
+    map_ll : int
+        The MAP(maximum a posteriori probability) logliklihood
+    map_transmat_ :  array_like, shape = (n_states_, n_states_)
+        The map transition matrix
+    map_population_ :  array_like, shape = (n_states_, n_states_)
+        The MAP population
 
     Notes
     -----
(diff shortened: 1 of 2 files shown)


mpharrigan:
Why did you remove my lovely checkbox for "test added" :)

msultan:
@mpharrigan and here i thought I was being all sneaky about this.
this is ready for code review btw

cxhernandez:
It failed on Travis and pass Appveyor...? There's something new...

msmbuilder/msm/bayesmsm.py
```diff
@@ -235,6 +247,15 @@ def _fit_reversible(self, countsmat):
         # we round chain_length UP, so we might have generated a few extra
         # samples.
         result = result[-self.n_samples:]
+
+        self.all_logliklihoods_ = [_logprob_T(i,Z) for i in result]
```
cxhernandez: `all_loglikelihoods_`?

msultan: damn it

msultan:
Anyone have strong opinions on whether we should report the mean(as it is now) or the median reversible estimator as the MAP estimate?

msultan:
Interestingly enough, the start of each iteration on each of my cpu cores seems to give the best mdl. That is not too surprising since i think its just returning the unperturbed tmat back 

image

msultan:
Also if someone has any papers about how to change this sampler so that it doesnt require a reversible counts matrix, I am all ears. Currently, it adds the transpose to the counts matrix which I think is major flaw since it gets rid of a lot of interesting structure within our data. It technically shouldn't matter when we have a lot of transitions but that is rarely the case for larger systems of interest.

msultan:
image
Here is the histogram for one of my datasets . Its pretty gaussian-y so mean/median should both be fine.

msultan:
@cxhernandez @mpharrigan anyone wanna read through the wiki page and make sure this is implemented correctly?

rmcgibbo:
this code doesn't make any sense to me. why are you taking the mean of the log posterior?

msultan:
@rmcgibbo I wanted the mode of the posterior distribution to report a single best model from the set of all mdls reported by the BayesianMSM. However, that would have required some binning. I thought it might be better to take the mean of the distribution under the assumption that it would be close to the mode given that the distribution is gaussian.

rmcgibbo:
the "map" is the _maximum_ of the posterior density.

msultan:
@rmcgibbo yea, i think i misunderstood that. I will delete this PR. what would be a good way to get the best representative model out from the Bayesian MSM.

msultan:
btw, thanks a lot for the catch @rmcgibbo . I completely misunderstood, what distribution I was supposed to use.

Issue #745, opened 2016-03-04
Bootstrap msm is slow 


msultan:
The code is annoyingly slow. I think what is happening is that solving the eigensystem on the individual processors is holding things up whereas something like the ```bayesmsm``` is returning the tmats which are then processed sequentially on the fly. 

Two possible solutions:
1). Forget the parallel and fit the bootstrap models sequentially. 
2). Keep the parallel and only return the tmat ala ```bayesmsm``` and calculate everything sequentially . 

Basically, I need to un-engineer this because I thought I would be clever with parallelization.

PR #756, opened 2016-03-13
fix bootstrap slowness 

msultan:
 - [ ] Add tests
 - [ ] Update changelog

Based upon what i saw on real data and raised in #745 . Current hypothesis is that that eigensolver is slowing things down. The ```_parallel_fit``` command should trigger the eigensolver sequentially any how as it tries to map the perturbed populations back to the mle solution.

### DIFF
--- a/msmbuilder/msm/validation/bootstrapmsm.py
+++ b/msmbuilder/msm/validation/bootstrapmsm.py
@@ -215,7 +215,6 @@ def _fit_one(jt):
     try:
         mdl.fit(sequences)
         # solve the eigensystem
-        mdl.eigenvalues_[0]
     except ValueError:
         mdl = None
         warnings.warn("One of the MSMs fitting "


cxhernandez:
can you use [`%prun`](http://scikit-learn.org/stable/developers/performance.html#profiling-python-code) to find out for sure?

msultan:
```
       411794 function calls (411782 primitive calls) in 32.834 seconds
Seems like fitting is still taking a while but the code is similar to the ```Bayesianmsm``` code so i am not sure where else we could shave time off. 
   Ordered by: internal time

   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
       17   16.569    0.975   16.569    0.975 {method 'acquire' of '_thread.lock' objects}
        6   11.701    1.950   11.701    1.950 {msmbuilder.msm._markovstatemodel._transmat_mle_prinz}
     3577    3.104    0.001    3.554    0.001 function_base.py:1869(_vectorize_call)
    20987    0.516    0.000    0.516    0.000 {built-in method numpy.core.multiarray.array}
      456    0.292    0.001    0.292    0.001 {method 'sort' of 'numpy.ndarray' objects}
     2577    0.122    0.000    0.122    0.000 {built-in method numpy.core.multiarray.concatenate}
        6    0.117    0.019    4.340    0.723 core.py:487(_transition_counts)
   300883    0.053    0.000    0.053    0.000 core.py:546(<genexpr>)
     3577    0.046    0.000    3.599    0.001 function_base.py:1783(__call__)
       30    0.041    0.001    0.041    0.001 {method 'dot' of 'numpy.ndarray' objects}
```

msultan:
Gonna merge this now.

msmbuilder/msm/validation/bootstrapmsm.py
```diff
@@ -215,7 +215,6 @@ def _fit_one(jt):
     try:
         mdl.fit(sequences)
         # solve the eigensystem
```
cxhernandez: get rid of this for clarity?

cxhernandez:
Perhaps we could be smarter about calculating the MSM multiple times? Maybe there's some overhead in just initializing `msmbuilder.msm._markovstatemodel._transmat_mle_prinz`, which we could shave off?

msultan:
I thought about implementing it that way first where a single MSM object was being passed to the workers each of whom would fit it to a different copy of the data. I was not sure if that was safe to do since I am not certain whether the object is immutable or not.

msmbuilder/msm/validation/bootstrapmsm.py
msultan: next pr?

cxhernandez: Sure. I was too slow for this one.

PR #757, opened 2016-03-15
MSM state visualization

jadeshi:
 - [ ] Add tests
 - [ ] Update changelog

[Describe changes here]

Hey all,
I'd like to share some code of mine for computing representative structures for MSM states. Basically, it computes a "centroid" structure for each state based on the featurization scheme, returning an xtc of the corresponding structures from the raw data. This is reminiscent of the Gens file from the legacy MSMBuilder 2 and useful for getting some visual insight into the MSM states. Most of the code is essentially a rehash of "draw_samples.py", and it's basically just an easy, straightforward implementation of it. This script takes three main input arguments, which should hopefully be straightforward to generate:

1. A text file of trajectory paths to the (raw, unfeaturized) simulation trajectories. The order of these trajectories in the file should be the order in which they were loaded when building your model. If you strided your data while loading it, you should pass that same stride here.

2. A .npy file of the featurized trajectory data, which will be used to compute the centroid frame for each state that will then be pulled from the corresponding raw trajectory.

3. A .npy file of "assignments", which maps all frames of the raw data onto the MSM state space. This is the .labels_ attribute of your featurizer object after you fit the raw data. 

I've used this with my own work and there don't seem to be any obvious bugs. Hopefully this is useful for people working with MSMs and makes the models more physically interpretable. Any comments / suggestions are welcome. 

Thanks! 
Jade


msultan:
what about also adding in an msm object so that it can account for discarding of states?

jadeshi:
I like that idea. I think an easier way may be to just add another argument that allows for custom selection of which states to pull from.

mpharrigan:
You're missing a comma somewhere

jadeshi:
Thanks for the catch. Fixed.

msmbuilder/utils/msmb3_centroid.py
```diff
@@ -0,0 +1,86 @@
...
+            output = output.join(raw_frame)
+    output = output.superpose(output)
+    return output
+
+
+if __name__ == '__main__':
+    parser = argparse.ArgumentParser(description='''Computes a representative structure for each state in an MSM and returns the structures, as a single xtc file. Used to get a visual interpretation of the MSM states, reminiscent of the "Gens" file in MSMBuilder 2.''')
+    parser.add_argument('-t', '--trj_paths', type=str, help='''A text file with raw trajectory paths, the order they were loaded while you were building the MSM.''')
+    parser.add_argument('-f', '--featurized_trjs', type=str, help='''Featurized trajectories file in .npy format.''')
+    parser.add_argument('-s', '--stride', type=str, help='''If you strided your data before featurizing it to build your MSM, use that corresopnding stride here.''', default='1')
```
cxhernandez: typo: corresopnding

jadeshi:
@cxhernandez Thanks! Fixed.

mpharrigan:
I worry about including these untested, one-off scripts that don't integrate tightly with the rest of the api into the main source tree. But I think it's important for this code to be shared.

What about putting scripts like this into a top level directory `contrib/` (or something). These scripts wouldn't have any backwards compatibility guarantees.

msultan:
@jadeshi if you have a 100 state mdl and 1 state gets discarded, would this require a person to type 1 through 99 manually in the command line?

jadeshi:
@msultan No, the script takes in a text file of state indices, which can straightforwardly be pulled from msm.mapping_

@mpharrigan I like that idea.

jadeshi:
Any idea why one of the tests failed? All I did was move the script to a newly created directory.

cxhernandez:
I'm thinking this is better suited for [msmexplorer](https://github.com/msmexplorer/msmexplorer/), esp once https://github.com/msmbuilder/msmbuilder/pull/768 is up and running

PR #751, opened 2016-03-09
fix figshare links

cxhernandez:
re: https://github.com/msmbuilder/msmbuilder/issues/750

PR #761, opened 2016-03-16
Backport figshare link update

mpharrigan:
Backport #751 to 3.3 branch for forthcoming 3.3.2


mpharrigan:
Too much has changed since 3.3, making this sort-of a headache. Let's just push out 3.4 as quickly as possible

Issue #732, opened 2016-02-16
StrucRMSDFeaturizer and RMSDFeaturizer do the same thing


mpharrigan:
https://github.com/msmbuilder/msmbuilder/blob/master/msmbuilder/featurizer/featurizer.py#L205
https://github.com/msmbuilder/msmbuilder/blob/master/msmbuilder/featurizer/featurizer.py#L892

cc #695 

Many apologies to @brookehus who was mislead into re-writing this featurizer :) Her code is probably better so we should use her code going forward.

I don't think we've made a release with `StrucRMSDFeaturizer` in it, so I propose throwing away existing `RMSDFeaturizer` and renaming `StrucRMSDFeaturizer` in its place (and adapting the order of arguments to make the api compatible)

mpharrigan:
@brookehus notes that she could not discover the rmsdfeaturizer because it didn't show up on the command line #733

brookehus:
I've addressed #734 on my own branch should I also address this one too with the same PR?

mpharrigan:
:+1:

mpharrigan:
This is somewhat important for the 3.4 release. We don't want to introduce a new api that will be removed shortly thereafter

brookehus:
I can finish this up and the command line stuff by Friday.

mpharrigan:
I think most of it is already there, so I'll finish this up

mpharrigan:
closed by #764

PR #764, opened 2016-03-16
Remove old RMSDFeaturizer code. Replace with what used to be StrucRMSDFeaturizer

mpharrigan:
Addresses #732


mpharrigan:
todo:
 - I don't think my new test(s) is actually running
 - rename test file to remove "struc" part

Issue #759, opened 2016-03-16
Add citation for sparseTICA in the docs


cxhernandez:
http://arxiv.org/abs/1602.08776

rmcgibbo:
+1

mpharrigan:
Closed by #765

PR #765, opened 2016-03-17
add sparse tica paragraph

cxhernandez:
https://github.com/msmbuilder/msmbuilder/issues/759


rmcgibbo:
the first sentence of this ("optimal natural reaction coordinate") really makes more sense in the context of the paper than copy-pasted here.

Issue #605, opened 2015-07-06
Implement tICA scaling


rmcgibbo:
see Noe and Clementi, 2015 (http://arxiv.org/abs/1506.06259)

kyleabeauchamp:
It's quite similar to the weighted transform option that we have, IIRC

rmcgibbo:
Yeah, it's not very different practically, but the theory is better.

mpharrigan:
Closed in #766

PR #766, opened 2016-03-18
change def of weighted transform

cxhernandez:
re: https://github.com/msmbuilder/msmbuilder/issues/605

### DIFF
--- a/msmbuilder/decomposition/tica.py
+++ b/msmbuilder/decomposition/tica.py
@@ -45,7 +45,7 @@ class tICA(BaseEstimator, TransformerMixin):
         Deprecated. Please use kinetic_mapping.
     kinetic_mapping : bool, default=False
         If True, weigh the projections by the tICA eigenvalues, yielding
-         kinetic distances.
+         kinetic distances as described in [6].
 
     Attributes
     ----------
@@ -97,6 +97,8 @@ class tICA(BaseEstimator, TransformerMixin):
     .. [4] Molgedey, Lutz, and Heinz Georg Schuster. Phys. Rev. Lett. 72.23
        (1994): 3634.
     .. [5] Chen, Yilun, Ami Wiesel, and Alfred O. Hero III. ICASSP (2009)
+    .. [6] Noe, F. and Clementi, C. arXiv arXiv:1506.06259 [physics.comp-ph]
+           (2015)
     """
 
     def __init__(self, n_components=None, lag_time=1, shrinkage=None,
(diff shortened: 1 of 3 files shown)


cxhernandez:
There's some random error in `msmbuilder.tests.test_ghmm.test_3_state` for appveyor

cxhernandez:
Also as a side note: I'm not sure this is as tiny a change as @rmcgibbo made it out to be. The convergence of my implied timescales improved greatly for my system after using this, but the slowest characteristic timescale was reduced by about 5-fold, which was way more than the 95% confidence interval around that value from bootstrapping. Currently running a bunch of osprey jobs to make sure this wasn't a fluke, but something to keep in mind.

mpharrigan:
How about we add a new option (`kinetic_distance` ?) and add a warning that `weighted_transform` is deprecated and people should use `kinetic_distance`. And add a test and update the changelog

cxhernandez:
more appveyor weirdness

msmbuilder/decomposition/tica.py
```diff
@@ -42,8 +42,10 @@ class tICA(BaseEstimator, TransformerMixin):
         specified (the default) it is estimated using an analytic formula
         (the Rao-Blackwellized Ledoit-Wolf estimator) introduced in [5].
     weighted_transform : bool, default=False
-        If True, weight the projections by the implied timescales, giving
-        a quantity that has units [Time].
+        Deprecated. Please use kinetic_mapping.
+    kinetic_mapping : bool, default=False
```
mpharrigan: Add a ref to frank's paper?

msmbuilder/decomposition/tica.py
```diff
@@ -321,7 +325,12 @@ def transform(self, sequences):
             X_transformed = np.dot(X, self.components_.T)
 
             if self.weighted_transform:
+                warnings.warn("weighted_transform is deprecated. "
+                              "Please use kinetic_mapping",
+                              DeprecationWarning)
                 X_transformed *= self.timescales_
+            elif self.kinetic_mapping:
```
mpharrigan: If both are set to `True`, can you raise an error?

msmbuilder/decomposition/tica.py
```diff
@@ -448,6 +457,7 @@ def summarize(self):
 shrinkage           : {shrinkage}
 lag_time            : {lag_time}
 weighted_transform  : {weighted_transform}
+kinetic_mapping : {kinetic_mapping}
```
mpharrigan: Can you space pad this so it's aligned like the other fields?

brookehus:
Don't mind me, just listening.

mpharrigan:
lgtm

PR #767, opened 2016-03-19
SparseTICA improvements

mpharrigan:
 - [ ] Add tests
 - [ ] Update changelog

I'm trying to just do sparse tica on the vonmises features for the example fs peptide data. I came across these bugs, but even with these fixes the heatmap looks positively whacky, so I suspect there's something else going on as well

### DIFF
--- a/msmbuilder/decomposition/sparsetica.py
+++ b/msmbuilder/decomposition/sparsetica.py
@@ -148,6 +148,11 @@ class SparseTICA(tICA):
             self._eigenvectors_[:, i] = v
             A = scdeflate(A, v)
 
+        # sort in order of decreasing value
+        ind = np.argsort(self._eigenvalues_)[::-1]
+        self._eigenvalues_ = self._eigenvalues_[ind]
+        self._eigenvectors_ = self._eigenvectors_[:, ind]
+
         self._is_dirty = False
 
     def summarize(self):
--- a/msmbuilder/decomposition/sparsetica.py
+++ b/msmbuilder/decomposition/sparsetica.py
@@ -112,10 +112,12 @@ class SparseTICA(tICA):
     """
 
     def __init__(self, n_components=None, lag_time=1, rho=0.01,
-                 weighted_transform=True, epsilon=1e-6, shrinkage=None,
+                 weighted_transform=False, kinetic_mapping=False,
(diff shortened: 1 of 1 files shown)


mpharrigan:
The lack of sorting is a pretty big deal, I think. So anyone who used this in production should probably review their work

mpharrigan:
Or is `speigh` supposed to sort automatically? In practice this doesn't seem to be happening

mpharrigan:
screenshot from 2016-03-19 16 46 05

 1. download fs peptide
 2. use von mises featurizer with default arguments
 3. transform with sparse tica

mpharrigan:
```
In [14]: t.eigenvalues_
Out[14]: array([ 0.99025034,  0.98690298,  0.98711887,  0.99172246,  0.98582245])
```

rmcgibbo:
Speigh is only a local optimizer that finds a single pseudo eigenvector starting from some initial guess, so you really can't be completely sure that you get them in sorted order.

cxhernandez:
good catch @mpharrigan!

mpharrigan:
@rmcgibbo, should we sort them after? should we set the number of components higher than what the user requests, sort, and then truncate?

rmcgibbo:
Sorting after is fine. I don't think you should set the number of
components higher than the user requests though. That's not guaranteed to
find the global max either. Better to just inform users that this is how it
works, and they can set the number of components themselves.

On Mon, Mar 21, 2016 at 2:31 AM, Matthew Harrigan <notifications@github.com>
wrote:

> @rmcgibbo <https://github.com/rmcgibbo>, should we sort them after?
> should we set the number of components higher than what the user requests,
> sort, and then truncate?
>
> —
> You are receiving this because you were mentioned.
> Reply to this email directly or view it on GitHub
> <https://github.com/msmbuilder/msmbuilder/pull/767#issuecomment-199145517>
>



-- 
-Robert

mpharrigan:
Ok, I added a note and we'll keep the sorting code (and addition of kinetic_map)

mpharrigan:
Still wondering why my fs-peptide heatmap is so wonky

mpharrigan:
Also removed the mention of cvxpy, which is no longer required

PR #769, opened 2016-03-20
Updated describe_features

msultan:
update all the describe_features!


mpharrigan:
Cool, let me know when this is ready for merge

mpharrigan:
Subsumes #677

msultan:
alright this is ready for code review. I had to add a test for the Vonmises featurizer to make sure its output was being put in the way I was expecting it. That should have been a separate PR. Sorry about that :/

msmbuilder/tests/test_featurizer.py
```diff
@@ -85,6 +86,19 @@ def test_von_mises_featurizer():
...
 
+    #test to make sure results are being put in the right order
+    feat = VonMisesFeaturizer(["phi","psi"], n_bins=10)
+    _, all_phi = md.compute_phi(trajectories[0])
+    X_all = feat.transform(trajectories)
+    all_res=[]
+    for frame in all_phi:
+        for dihedral_value in frame:
+            all_res.extend(vm.pdf(dihedral_value,loc=feat.loc, kappa=feat.kappa))
+    for k in range(10):
```
msultan: this checks 10 random features to make sure that they appear in the right columns

mpharrigan:
Tests are always welcome. I'll take a look

msmbuilder/tests/test_featurizer.py
mpharrigan: put that in a comment in the code?

msmbuilder/featurizer/featurizer.py
```diff
@@ -403,42 +405,64 @@ def __init__(self, types=['phi', 'psi'], sincos=True):
...
+                - featurizer: Dihedral
+                - featuregroup: the type of dihedral angle and whether sin or
+                  cos has been applied.
+        """
 
+        feature_descs = []
+        for dihed_type in self.types:
+            # TODO: Don't recompute dihedrals, just get the indices
+            func = getattr(md, 'compute_%s' % dihed_type)
+            # ainds is a list of four-tuples of atoms participating
```
mpharrigan: This comment references a variable that doesn't exist

msmbuilder/featurizer/featurizer.py
```diff
@@ -513,37 +537,66 @@ def __init__(self, types=['phi', 'psi'], n_bins=18, kappa=20.):
...
+        feature_descs : list of dict
+            Dictionary describing each feature with the following information
+            about the atoms participating in each dihedral
+                - resnames: unique names of residues
+                - atominds: the four atom indicies
+                - resseqs: unique residue sequence ids (not necessarily
+                  0-indexed)
+                - resids: unique residue ids (0-indexed)
+                - featurizer: Dihedral
+                - featuregroup: The bin index(0..nbins-1)
```
mpharrigan: Shouldn't "bin" be an additional field, or no? So one column is phi-psi and another column is bin. Whatever you think is best here for your use cases.

msmbuilder/featurizer/featurizer.py
```diff
@@ -513,37 +537,66 @@ def __init__(self, types=['phi', 'psi'], n_bins=18, kappa=20.):
...
+                    resid = set(top.atom(ai).residue.index for ai in ainds)
+                    all_aind.append(ainds)
+                    bin_info += ["vm%s"%bin_index]
+                    resids += [list(resid)]
+                    resseqs += [[top.residue(ri).resSeq for ri in resid]]
+                    resnames += [[top.residue(ri).name for ri in resid]]
+
+            zippy = zip(bin_info, all_aind, resseqs, resids, resnames)
+            #fast check to make sure we have the right number of features
+            assert len(bin_info)==len(aind_tuples) * self.n_bins
```
mpharrigan: pep8 spaces around `==` :8ball:

msmbuilder/featurizer/featurizer.py
```diff
@@ -629,44 +682,66 @@ def partial_transform(self, traj):
...
+                - resids: unique residue ids (0-indexed)
+                - featurizer: Alpha Angle
+                - featuregroup: the type of dihedral angle and whether sin or
+                  cos has been applied.
+        """
+        feature_descs = []
         # fill in the atom indices using just the first frame
         self.partial_transform(traj[0])
+        top = traj.topology
         if self.atom_indices is not None:
```
mpharrigan: This `if` block is somewhat annoying stylistically. It means the whole body of the function has to be indented.

How about
```python
if self.atom_indices is None:
  raise ValueError()
# do the body of the function here
# no need for an else!
```

msmbuilder/featurizer/featurizer.py
```diff
@@ -629,44 +682,66 @@ def partial_transform(self, traj):
...
+                    resseqs=resseq,
+                    resids=resid,
+                    featurizer="AlphaAngle",
+                    featuregroup="{}".format(sincos),
+                )]
+
+            return feature_descs
         else:
-            raise UserWarning("Cannot describe features for trajectories with fewer than 4 alpha carbon\
+            raise UserWarning("Cannot describe features for trajectories with "
```
mpharrigan: Don't "raise" a warning. This should probably be something like
```python
raise ValueError("...")
```

mpharrigan: And since it gets raised if self.atom_indices is None, you should probably include that in the error message

msmbuilder/featurizer/featurizer.py
```diff
@@ -629,44 +682,66 @@ def partial_transform(self, traj):
...
+                              "fewer than 4 alpha carbon\
```
mpharrigan: Don't use `\` line continuation in strings, it will include all the spaces! Use auto-string concat
```python
func("a long string "
        "that spans multiple lines.")

File: msmbuilder/featurizer/featurizer.py [hunk 8]
```diff
@@ -837,33 +938,60 @@ def partial_transform(self, traj):
...
+                - resnames: unique names of residues
+                - atominds: the four atom indicies
+                - resseqs: unique residue sequence ids (not necessarily
+                  0-indexed)
+                - resids: unique residue ids (0-indexed)
+                - featurizer: Contact
+                - featuregroup: ca, heavy etc.
+        """
+        feature_descs = []
         # fill in the atom indices using just the first frame
```
mpharrigan: This uses all the frames

msmbuilder/featurizer/featurizer.py
```diff
@@ -837,33 +938,60 @@ def partial_transform(self, traj):
...
+        aind = []
+        resseqs = []
+        resnames = []
+        for resid_ids in residue_indices:
+            aind += ["N/A"]
+            resseqs += [[top.residue(ri).resSeq for ri in resid_ids]]
+            resnames += [[top.residue(ri).name for ri in resid_ids]]
+
+
+        zippy = zip(aind, resseqs, residue_indices, resnames)
```
mpharrigan: This immediately gets overwritten

msmbuilder/featurizer/featurizer.py
msultan: havent used this featurizer but i was assuming we wanted similar column names. @cxhernandez you wanna chime in here?

msmbuilder/featurizer/featurizer.py
msultan: i think the other error is more descriptive instead of a generic atom_indices error. I will raise the error as a ValueError explicitly though.

msultan:
hmm, there is a bunch of repeating code here, in particular the ```aind_tuples``` for loop, that should probably get shoved into a helper routine for the featurizer base class.

mpharrigan:
> hmm, there is a bunch of repeating code here, in particular the aind_tuples for loop, that should probably get shoved into a helper routine for the featurizer base class. 

I was thinking that too. So many different variations on computing a dihedral angle!

msultan:
alright, i changed a bunch of the code and made the vonmises output similar to that of other featurizers. @cxhernandez could you test the Vonmises ? @mpharrigan does this refactoring work?

mpharrigan:
I don't like how dict_maker takes a list and appends to it. "Side-effects" like these will cause confusion later (or now when I was trying to figure out what the two arguments to the helper function are) and hurt composability.

Have dict_maker return a list of dicts. The caller will handler the concatenation of many of these calls

mpharrigan:
Does this change the order of the features for vonmises? Not an issue if "yes", just curious

msultan:
```quote
Does this change the order of the features for vonmises? Not an issue if "yes", just curious```

yea, i changed it from dihedral->bins to bins-> dihededrals. it more consistent with the sin, cos transformations that other featurizers use.

**Comment by msultan** [comment]
```quote
I don't like how dict_maker takes a list and appends to it. "Side-effects" like these will cause confusion later (or now when I was trying to figure out what the two arguments to the helper function are) and hurt composability.
```

Alright i will change that.

msultan:
alright, i think this is ready now. I have added tests to the feature_descriptors for most of them. 
The tests all follow the same basic pattern of

1. Pick a random trajectory
2. Transform it 
3. describe its features
4. pick 25 random features and re calculate them using mdtraj's base function classes. 
5. compare and assert for equivalence 

One of the things that popped up because of the tests was that the Kappa angle featurizer was performing the sin cos transform in the opposite order to the Dihedral featurizer. I have it fixed now.

msultan:
hold on, i forgot to add the vonmises test. adding it now

mpharrigan:
Nice. You should add yourself as "contributor" in the top of featurizer.py

msultan:
fantastic , windows is failing because of memory issues when trying to load the fs_peptide dataset.

mpharrigan:
Lol, That's really annoying. You can load a subset of them or stride them

```python
fs = FsPeptide()
fs.cache()
dirname = fs.data_dir
top = md.load(dirname+"/fs-peptide.pdb")
trajectories = [md.load(fn, stride=100) for fn in glob.glob(dirname + "trajectory-*.xtc", top=top)]
```

cxhernandez:
Is this good to merge? I think it's the last missing piece for the 3.4 release.

msultan:
oh this finally ran without erroring on windows

msultan:
@cxhernandez merge away.

mpharrigan:
I really don't like that it doesn't let you merge if it's not "up to date" with master. That's the whole g.d. point of a merge

msultan:
this is out of sync with master again. will merge as soon as travis passes.

msultan:
merging without all green checks...tsk tsk @mpharrigan

PR #768, opened 2016-03-20
Project Templates

mpharrigan:
This is a serious wip, but it's solid work towards making msmbuilder actually usable.

The general idea is we will develop a set of short scripts that do common tasks, the user will run `msmb SetUpProject`, and it'll dump a variety of scripts that will serve as a good starting point.

This PR also includes a new dataset idiom including a persistent pandas dataframe that keeps a mapping from a consistent set of keys (the dataframe index) to trajectory filenames and other metadata (topology filename, number of frames, whatever you want ... the user is encouraged to modify the "generate-metadata" script)

Data is still saved as a bunch of numpy files in a directory, but now instead of trying to walk the filesystem to find files, we use the dataframe to find everything we're looking for.

My grand plan is to get these scripts pretty solid for a "generic" project where trajectories are keyed on integers, and then start including fah-specific scripts [primarily modifying the metadata part] that set up a multi-indexed dataframe (proj,run,clone) and saves trajectory files in a proj/run/clone.npy hierarchy on the filesystem.

This PR ropes in #747 with modifications.

I think the existing dataset infrastructure is overkill in terms of lazy-loading. The amount of GB of data is low for everything besides featurization, so we can shove our whole dataset into memory.

For example, 1 millisecond of data saved every 1 ns with 1000 features as 32-bit floats is only 4GB. That being said, we'll still have to do the initial featurization in a memory-efficient manner

One of my grander plans is to change from treating "lists of timeseries-arrays" as the thing that flows through our estimators to "dictionaries of timeseries-arrays". I imagine modifying our `transform` wrapper to return a list if a list is given and a dictionary if a dictionary is given, for backwards compatibility.

1. the trajectories are inherently not in any order, as implied by a list
2. meaningful keys are often non-sequential. This puts undue burden on the user to keep track of a mapping from "real keys" to list indices
3. people want non-integer keys. For FAH-style projects, keying by (proj,run,clone) would be nice. Strings could be used to label trajectories "apo" or "bound".

For now, the scripts will include
```python
out = {}
for k, traj in trajs:
  out[k] = estimator.partial_transform(traj)
```

But it would be nice to just do
```python
out = estimator.transform(trajs)
```

PR #770, opened 2016-03-23
added ability to normalize large datasets in memory

evanfeinberg:
This PR adds a "preprocessing" section which is inspired by the approach of geometric clustering in `MSMBuilder`: take standard `sklearn` methods and expand them to handle `list`s of `np.array`s. This is more convenient, as nearly all `sklearn` methods assume that `X` is a single `np.array` whereas MD data is mostly stored in lists of arrays. 

Large featurized [MD] datasets are often large compared to total RAM available on a system. If size of the MD dataset is comparable to the total RAM available, then computing the mean and variance with standard `numpy` functions can be tricky. Since concatenating the data essentially doubles the RAM that needs to be used, the usual way of computing mean and variance can in turn make the user exceed the memory limits of the system.

To begin, I implemented a numerically stable method described in https://en.wikipedia.org/wiki/Algorithms_for_calculating_variance#Compensated_variant. In addition to the normal `fit`, `transform`, and `fit_transform` methods, I added `fit_transform_in_place` and `transform_in_place`, so one can pass in un-normalized features and compute normalized features without adding to the memory burden of one's computation. 

Some to-do's: add whitening and other preprocessing capabilities from `sklearn` to `msmbuilder.preprocessing`. 

Comments or criticisms are welcome. Thank you to @cxhernandez  in particular for his help with this PR.


mpharrigan:
What is `msmbuilder/preprocessing/._normalize.py`? Can you remove it and rebase so it doesn't take up space in the history

msmbuilder/preprocessing/base.py
```diff
@@ -0,0 +1,126 @@
+# Author: Matthew Harrigan <matthew.p.harrigan@gmail.com>
```
mpharrigan: lol I don't *think* I wrote this :)

msmbuilder/preprocessing/normalize.py
```diff
@@ -0,0 +1,157 @@
+# Author: Evan N. Feinberg <efein19@gmail.com>
+# Contributors: Christian Schwantes <schwancr@gmail.com>m Robert McGibbon <rmcgibbo@gmail.com>, Kyle A. Beauchamp  <kyleabeauchamp@gmail.com>
```
mpharrigan: Hard-wrap this line

mpharrigan:
I don't really understand this PR. The end goal is to remove mean and normalize variance for feature-trajectories (aka 2d numpy arrays)? Can you really not fit a featurized trajectory in ram?

mpharrigan:
Back of the envelope, one milisecond saved every nanosecond with 1000 features only takes 4gib of ram. This is the same example I used in #768 to argue for just loading everything into ram

This is also pretty conservative. For normalizing, you can load in one trajectory at a time (using `dataset`) which will be << 1 ms or save less frequently than every 1 ns

mpharrigan:
Oh I guess you have to consider *all* the data when doing mean+std calculations

msmbuilder/preprocessing/normalize.py
```diff
@@ -0,0 +1,157 @@
...
+        `partial_fit()` updates the fit with new data, and is suitable for
+         online learning.
+
+    Note: Numerically stable method of computing variance adapted from:
+    https://en.wikipedia.org/wiki/Algorithms_for_calculating_variance
+    """
+
+    def __init__(self):
+
+        """
```
mpharrigan: `__ini__` generally doesn't have a docstring. Put this in the class docstring or as line comments `# (these)`

msmbuilder/preprocessing/normalize.py
```diff
@@ -0,0 +1,157 @@
...
+
+        self.k = len(sequences)
+        self.n = 0.
+        self.sum1 = 0.
+        self.sum2 = 0.
+        self.sum3 = 0.
+
+        #First Pass
+        for sequence in sequences:
+            self.n += float(np.shape(sequence)[0])
```
mpharrigan: Why is `n` a float?

msmbuilder/preprocessing/normalize.py
mpharrigan: I also suspect not all these people contributed

mpharrigan:
Why do you include ` msmbuilder/preprocessing/base.py`?

mpharrigan:
Don't you need to implement `partial_transform()` in the Normalize class?

mpharrigan:
This also needs a suite of tests :)

msmbuilder/preprocessing/base.py
evanfeinberg: You indeed wrote the vast majority of it!

evanfeinberg:
Yes this is precisely why this PR is necessary. Suppose your data, `sequence` takes up 10 GB in memory and your system only has 16 GB of RAM (both very sensible figures). Computing mean and (especially) variance requires seeing all the data. Calling `np.concatenate` on `sequence` is essentially a copy. If you want to keep around `sequence` and `sequence_concatenated`, this will push memory usage beyond the 16 GB limit. In addition, for transforming, making a `sequence_normalized` separate from `sequence` would similarly ~double memory requirements. Therefore a `transform_in_place` is useful.

evanfeinberg:
Doesn't `msmbuilder/preprocessing/base.py` obviate implementing `partial_transform` ?

msmbuilder/preprocessing/normalize.py
evanfeinberg: Agreed, will be removed.

msmbuilder/preprocessing/normalize.py
evanfeinberg: So when we divide sum of squares by `n` Python doesn't get confused and cast the sum to an `int` which would be bad. Unless it automatically casts `n` to a float underneath the hood in which case we're ok?

evanfeinberg:
Also how can I add tests?

msmbuilder/preprocessing/normalize.py
evanfeinberg: Oops you're totally right, corrected.

msmbuilder/preprocessing/normalize.py
mpharrigan: python2 will truncate it to int (dumb python 2!). python3 will be nice. You can get the python3 behavior by putting

```python
from __future__ import division
```

at the top

mpharrigan:
> Also how can I add tests?

Check out the files in `msmbuilder/tests/`.

I would recommend adding a new file called `test_normalize.py`. In it, you can put a couple functions whose name start with `test_` that each do one little thing and make an assertion at the end. These functions will all be run on travis, and if anything raises an exception or assertion error, that particular test will fail. You can use the script `runtests.py` in the project root to re-build and run the tests. You can also run `nosetests msmbuilder` to test your currently installed copy. To just do a particular test file, you can do `nosetests msmbuilder.tests.test_normalize`. You can't run this program from the source directory! or it will try to import the local files instead of the built+installed copy

mpharrigan:
> Doesn't msmbuilder/preprocessing/base.py obviate implementing partial_transform ?

You don't inherit from any of the classes there. As it stands, I don't see you using any of the functionality form base.py. If that's just a copy-paste of the file in decomposition... don't. import from decomposition. But I'm pretty sure you're not using and shouldn't use any of that functionality. The whole point of this PR is to avoid concatenating things

mpharrigan:
> Suppose your data, sequence takes up 10 GB in memory and your system only has 16 GB of RAM

Can you address my envelop calculation

> one milisecond saved every nanosecond with 1000 features only takes 4gib of ram

evanfeinberg:
Wow, WIP #768 looks really cool! Unfortunately, if you have several thousand features, as I often do, 1000 features / 4GB --> 5000 features / 20 GB easily. For instance, suppose one has 1 ms of trajectory subsampled at every 1 ns. Measuring ~2500 `closest-heavy` and `CA` distances translates to 5000 features and 20 GB of data. On a lot of hardware this is a nightmare.

evanfeinberg:
With regards to base.py both `transform()` and `fit_transform()` are only implemented in base.py, not in normalize.py. I figured having base.py in the class would be useful to other `preprocessing` classes we will end up having. But yes right now it is mostly a pared-down copy paste.

msmbuilder/preprocessing/normalize.py
evanfeinberg: cool, done!

mpharrigan:
> With regards to base.py both transform() and fit_transform() are only implemented in base.py, not in normalize.py.

You don't inherit from it anywhere. The code in base.py isn't used

rmcgibbo:
I just saw this on my phone and didn't read rhe whole thing... but IIRC in
mdtraj I implemented running estimators of the mean and variance that take
a single pass through the data and are in C or cython so they should be
both fast and stable. Could be useful for this PR.

rmcgibbo:
Here's the code in mdtraj for an online (running) calculation of the first couple moments of a bunch of data. It doesn't look like I've wrapped this from cython, but if it would be useful I could do that.

https://github.com/mdtraj/mdtraj/blob/master/mdtraj/geometry/src/moments.cpp

evanfeinberg:
Oh cool, it looks like I ended up duplicating the same algorithm implementation from that great Wiki page on variance computation. @rmcgibbo is this function callable from mdtraj Python API easily?

cxhernandez:
> It doesn't look like I've wrapped this from cython, but if it would be useful I could do that.

^ @evanfeinberg

msultan:
can we get this into master? Seems like a very useful addition

mpharrigan:
There are still a couple of issues with the code that need to be addressed

1. docstrings
2. completely extraneous file that I think @evanfeinberg thinks he is using but actually is not
3. tests

And introducing a whole new API not found in sklearn is something I would like to give some thought to.

1. is this the best name? can we do this as an init arg instead?
2. how does sklearn handle these issues? can we do something like that?

It also looks like there was a conversation about using the c implementation to make this speedy, so @evanfeinberg and @rmcgibbo need to decide whether they're going to do that

mpharrigan:
For the second point 2, perhaps interested parties can look into this

http://scikit-learn.org/stable/modules/preprocessing.html

http://scikit-learn.org/stable/modules/generated/sklearn.preprocessing.StandardScaler.html#sklearn.preprocessing.StandardScaler

I'd be more comfortable using a thin wrapper around sklearn. At the very least, we should copy their api design. It's essentially the second part of my second number 1: they use an init arg `copy`. When set to false, it does stuff in place

cxhernandez:
I've implemented point 2 in https://github.com/msmbuilder/msmbuilder/pull/796.

@evanfeinberg: I still think it's worth checking if @rmcgibbo's code runs faster than the sklearn version.

mpharrigan:
I'm going to close this because of #796 . Feel free to file additional PRs for performance enhancements if necessary

PR #788, opened 2016-04-01
added appreciation for Andrey Markov

msultan:
I noticed that msmbuilder currently doesn't appreciate [Andrey Markov](https://en.wikipedia.org/wiki/Andrey_Markov) enough. This PR attempts to fix this by adding a markov_appreciation module to msmbuilder. 

Currently only works for py3.

### DIFF
--- a/msmbuilder/msm/markov_appreciation.py
+++ b/msmbuilder/msm/markov_appreciation.py
@@ -1,0 +1,17 @@
+# Author: Muneeb Sultan <msultan@stanford.edu>
+# Contributors:
+# Copyright (c) 2014, Stanford University
+# All rights reserved.
+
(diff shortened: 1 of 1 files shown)


cxhernandez:
I think you meant to include this image:
image

msmbuilder/msm/markov_appreciation.py
```diff
@@ -0,0 +1,17 @@
+# Author: Muneeb Sultan <msultan@stanford.edu>
+# Contributors:
+# Copyright (c) 2014, Stanford University
```
mpharrigan: Please update this date

cxhernandez:
Also plz add a test

PR #796, opened 2016-04-10
port sklearn preprocessing

cxhernandez:
 - [x] Add tests
 - [x] Update changelog

re: https://github.com/msmbuilder/msmbuilder/pull/770#issuecomment-207139793

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -18,6 +18,12 @@ New Features
  - ``SparsePCA`` and ``MiniBatchSparsePCA`` are available in the
    ``decomposition`` module (#791).
 
+ - ``Binarizer``, ``FunctionTransformer``, ``Imputer``, ``KernelCenterer``,
+   ``LabelBinarizer``, ``MultiLabelBinarizer``, ``MinMaxScaler``,
+   ``MaxAbsScaler``, ``Normalizer``, ``RobustScaler``, ``StandardScaler``,
+   and ``PolynomialFeatures`` are available in the ``preprocessing``
+   module (#796).
+
 
 Improvements
 ~~~~~~~~~~~~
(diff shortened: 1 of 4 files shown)


cxhernandez:
I couldn't get `LabelEncoder` or `OneHotEncoder` to work properly. I'm not sure they're that useful for our purposed though....

cxhernandez:
@mpharrigan: any idea how to fix py34?

cxhernandez:
Might have to make two versions of `MinMaxScaler` and `StandardScaler`, since `partial_fit` was not implemented in these until sklearn v0.17. Should I call the newer ones `MinMaxOnlineScaler` and  `StandardOnlineScaler`?

mpharrigan:
> Might have to make two versions of MinMaxScaler and StandardScaler, since partial_fit was not implemented in these until sklearn v0.17. Should I call the newer ones MinMaxOnlineScaler and StandardOnlineScaler along with a deprecation warning?

-1 from me. Can we just use the new one and give a helpful error message if someone tries to use it with insufficient scipy

mpharrigan:
Also this pr is fantastic! Thanks for doing this

cxhernandez:
I'm just putting in conditionals so that if it doesn't exist in sklearn, the class won't be generated. Hopefully by the time msmb v3.5 is out, sklearn v0.17 is available for all platforms.

cxhernandez:
@mpharrigan: approval to merge?

mpharrigan:
Looks great from my end. I'm going to turn off the required status checks so you can merge

mpharrigan:
Actually I'm not going to do that because they added an option to turn off the requirement that the branch be up to date, which was my huge gripe. I'm going to leave the ci requirements (for now) such that we are motivated to fix the random test failures

cxhernandez:
> motivated to fix the random test failures

😞

Issue #790, opened 2016-04-06
[FR] Add possibility to featurized by binary "contacts" defined by a cutoff.


dr-nate:
In the MSM field, there might be some confusion regarding the difference between the term "contacts" versus "distances", where

```
Distances = distance between atoms
Contacts = binary 0/1 value based upon whether an atom-atom distance was below a cutoff
```
The benefit of using contacts is that you can often use more features, as the 0/1 requires less memory, and it is less sensitive to slight changes in the system. This will also help us match up with the *de novo* structure prediction people who define it the same way, where a search of the PDB yielded 8Å as the magic cutoff for determining significant contacts for protein CAs, for example. 

Currently, the `ContactsFeaturizer` only supports distances. Implementing the distinction could possibly be done by either adding a `cutoff` argument to the `ContactsFeaturizer` as it is, or by having separate featurizers for the two.

This shouldn't require changes to any other parts of the code, as I believe tICA will work fine on either type of value.

mpharrigan:
To realize the memory savings, we'd have to be careful with our numpy datatypes

cxhernandez:
Do you think a logistic version of this could be useful as well ?

Trayder:
This would be useful, I've been using binary contacts for a while as it has advantages for multi-body systems. And although its not much effort to convert a featurization into "binary", the current workflow does not play nice with integers so my 1's and 0's have been floats.

jchodera:
Is tICA particularly efficient or scalable when the contacts are binary
beyond just the memory savings?

dr-nate:
We used it pretty extensively in Gianni's group, and I believe Frank's
group does as well. I can't remember a time when we didn't have the
capability to do contacts+tica, and it had clear advantages with certain
application types like ligand binding. Perhaps distances would be
superior given more preprocessing like regularization, but that remains to
be shown IMO.

On Friday, April 8, 2016, John Chodera <notifications@github.com> wrote:

> Is tICA particularly efficient or scalable when the contacts are binary
> beyond just the memory savings?
>
> —
> You are receiving this because you authored the thread.
> Reply to this email directly or view it on GitHub
> <https://github.com/msmbuilder/msmbuilder/issues/790#issuecomment-207455375>
>


-- 
Sent from my phone.

mpharrigan:
> the current workflow does not play nice with integers

Can you elaborate?

mpharrigan:
[...]

PR #798, opened 2016-04-10
cutoff option for contact featurizer

brookehus:
 - [x] Add tests
 - [x] Update changelog

Attempts to address #790 although not in a memory-saving way.

Simple code to incorporate mapping contacts to 1/0 if they are under/over a certain cutoff. They are temporarily cast as floats... the code is:
`distances = (distances < self.binary_cutoff).astype(float)`

Obviously can change the type as we want.

Also, I couldn't find in the [docs](http://mdtraj.org/latest/api/generated/mdtraj.compute_contacts.htmll) what the distances units are.

### DIFF
--- a/msmbuilder/cluster/kcenters.py
+++ b/msmbuilder/cluster/kcenters.py
@@ -1,6 +1,6 @@
 # Author: Robert McGibbon <rmcgibbo@gmail.com>
-# Contributors:
-# Copyright (c) 2014, Stanford University
+# Contributors: Brooke Husic <brookehusic@gmail.com>
+# Copyright (c) 2016, Stanford University
 # All rights reserved.
 
 #-----------------------------------------------------------------------------
@@ -77,6 +77,9 @@ class _KCenters(ClusterMixin, TransformerMixin):
         self.random_state = random_state
 
     def fit(self, X, y=None):
+        if isinstance(X, np.ndarray):
+            if not (X.dtype == 'float32' or X.dtype == 'float64'):
+                X = X.astype('float64')
         n_samples = len(X)
         new_center_index = check_random_state(self.random_state).randint(0, n_samples)
 
@@ -115,6 +118,9 @@ class _KCenters(ClusterMixin, TransformerMixin):
         Y : array, shape [n_samples,]
             Index of the closest center each sample belongs to.
         """
+        if isinstance(X, np.ndarray):
+            if not (X.dtype == 'float32' or X.dtype == 'float64'):
+                X = X.astype('float64')
         labels, inertia = libdistance.assign_nearest(
             X, self.cluster_centers_, metric=self.metric)
         return labels
--- a/msmbuilder/commands/featurizer.py
+++ b/msmbuilder/commands/featurizer.py
@@ -13,7 +13,8 @@ from ..featurizer import (AtomPairsFeaturizer, SuperposeFeaturizer,
                           DRIDFeaturizer, DihedralFeaturizer,
                           ContactFeaturizer, GaussianSolventFeaturizer,
                           KappaAngleFeaturizer, AlphaAngleFeaturizer,
-                          RMSDFeaturizer)
+                          RMSDFeaturizer, BinaryContactFeaturizer,
+                          LogisticContactFeaturizer)
 
 
 class FeaturizerCommand(NumpydocClassCommand):
@@ -72,7 +73,7 @@ class FeaturizerCommand(NumpydocClassCommand):
         print("  >>> from msmbuilder.dataset import dataset")
         print("  >>> ds = dataset('%s')\n" % self.transformed)
 
-        if self.out is not '':
+        if self.out != '':
             verbosedump(self.instance, self.out)
             print("To load this %s object interactively inside an IPython\n"
                   "shell or notebook, run: \n" % self.klass.__name__)
@@ -164,7 +165,27 @@ class ContactFeaturizerCommand(FeaturizerCommand):
     klass = ContactFeaturizer
 
     def _contacts_type(self, val):
-        if val is 'all':
+        if val == 'all':
+            return val
+        else:
+            return np.loadtxt(val, dtype=int, ndmin=2)
+
+class BinaryContactFeaturizerCommand(FeaturizerCommand):
+    _concrete = True
+    klass = BinaryContactFeaturizer
+
+    def _contacts_type(self, val):
+        if val == 'all':
+            return val
+        else:
+            return np.loadtxt(val, dtype=int, ndmin=2)
+
+class LogisticContactFeaturizerCommand(FeaturizerCommand):
+    _concrete = True
+    klass = LogisticContactFeaturizer
+
+    def _contacts_type(self, val):
+        if val == 'all':
             return val
         else:
             return np.loadtxt(val, dtype=int, ndmin=2)
--- a/msmbuilder/featurizer/featurizer.py
+++ b/msmbuilder/featurizer/featurizer.py
@@ -853,7 +853,7 @@ class SASAFeaturizer(Featurizer):
 
 
 class ContactFeaturizer(Featurizer):
-    """Featurizer based on residue-residue distances
+    """Featurizer based on residue-residue distances.
 
     This featurizer transforms a dataset containing MD trajectories into
     a vector dataset by representing each frame in each of the MD trajectories
@@ -891,10 +891,12 @@ class ContactFeaturizer(Featurizer):
         self.scheme = scheme
         self.ignore_nonprotein = ignore_nonprotein
 
+    def _transform(self, distances):
+        return distances
 
     def partial_transform(self, traj):
-        """Featurize an MD trajectory into a vector space via of residue-residue
-        distances
+        """Featurize an MD trajectory into a vector space derived from
+        residue-residue distances
 
         Parameters
         ----------
@@ -913,9 +915,10 @@ class ContactFeaturizer(Featurizer):
         --------
         transform : simultaneously featurize a collection of MD trajectories
         """
+
         distances, _ = md.compute_contacts(traj, self.contacts,
                                            self.scheme, self.ignore_nonprotein)
-        return distances
+        return self._transform(distances)
 
 
     def describe_features(self, traj):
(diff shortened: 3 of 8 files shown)


mpharrigan:
mdtraj always turns everything into nanometers

(http://docs.openmm.org/6.3.0/userguide/theory.html#units for a brief discussion of why this is a good idea (tm))

mpharrigan:
+1 on this pr. some questions:

Why do you cast to float? If you just let it return numpy bools, it would save memory and probably be fine

Add a test(s) :)

We should consider (for usability's sake) having a "suggested" cutoff. I think nate said 8A. There are a couple ways I think we could do this

(a) accept `binary_cutoff='on'`, which means 8 angstroms or whatever we think is best in addition to a float. This is similar to what we do for MSM's `ergodic_cutoff`. We should accept `off` as well if we do this

(b) really I think people expect "contacts" to be binary. We can't change the behavior of this existing featurizer or we might screw up some people's calculations. What we could do is create two featurizers: `BinaryContactsFeaturizer` and `ContactDistanceFeaturizer`, and have the `ContactFeaturizer` be deprecated and say to pick one of those two. If we're patient enough, we can go through a full cycle of [deprecate, remove, re-introduce as binary contacts]

cxhernandez:
> What we could do is create two featurizers: BinaryContactsFeaturizer and ContactDistanceFeaturizer

+1. Also, add a logistic one if it's not too much trouble!

brookehus:
> Why do you cast to float? If you just let it return numpy bools, it would save memory and probably be fine

Is this because tica, clustering, etc. casts to whatever it needs?

And agree, a hybrid, deprecated ContactFeaturizer and then two new featurizers that are unambiguous sound good. I'll incorporate into the PR. @cxhernandez, what's the logistic one supposed to do exactly?

cxhernandez:
Basically the function would just transform the distances using:



where *L* is 1, Xo is the distance cutoff, and k is a tunable parameter which controls the ''softness" of the binning.

brookehus:
What would k be initialized to? Of course it would get thrown into osprey eventually, but is there any precedent?

cxhernandez:
> What would k be initialized to?

Not sure. You would definitely have to play around with it, but my guess would be somewhere around 2 inverse angströms? That way you get from 0 to 1 within half an angstrom. 

Could also be heuristically set as a fraction of the cutoff.

brookehus:
@mpharrigan: in a case where the booleans output from the binary contact featurizer aren't converted to floats, if we don't use tICA and go straight to clustering, e.g. minibatchkmeans, we get an output like:

```
/Users/brooke/anaconda/lib/python2.7/site-packages/sklearn/cluster/k_means_.py:1433: RuntimeWarning: Got data type bool, converted to float to avoid overflows
  X = self._check_test_data(X)
```

Therefore, to me, it seems preferable to convert to float within the featurizer. What do you think?

cxhernandez:
Woud `uint8 ` be a compromise?

brookehus:
I think it wants a float. `float16`?

mpharrigan:
I don't see the problem. kmeans apparently wants float for a legitimate reason, and they cast to float. no problem.

msmbuilder/featurizer/featurizer.py
```diff
@@ -851,8 +851,132 @@ def __init__(self, mode='residue', **kwargs):
     def partial_transform(self, traj):
         return md.shrake_rupley(traj, mode=self.mode, **self.kwargs)
 
-
+# BEH: normal
```
mpharrigan: ?

msmbuilder/featurizer/featurizer.py
```diff
@@ -851,8 +851,132 @@ def __init__(self, mode='residue', **kwargs):
...
 class ContactFeaturizer(Featurizer):
+    """Deprecated. Please use BinaryContactFeaturizer or
```
mpharrigan: The intro sentence should be one line

mpharrigan:
Looks ok to me.

1. can we reduce code duplication?

2. can you put the new argument (binary_cutoff) last, to retain backwards compatibility when using positional arguments

brookehus:
@mpharrigan @cxhernandez redone with a base class for distances and two classes that inherit it (binary and logistic), with a test, and with a few lines in some clustering codes to make sure things are cast as floats first, since that is what libdistance requires.

msmbuilder/commands/featurizer.py
```diff
@@ -169,6 +170,26 @@ def _contacts_type(self, val):
         else:
             return np.loadtxt(val, dtype=int, ndmin=2)
 
+class BinaryContactFeaturizerCommand(FeaturizerCommand):
+    _concrete = True
+    klass = BinaryContactFeaturizer
+
+    def _contacts_type(self, val):
+        if val is 'all':
```
mpharrigan: this needs to use `==`

mpharrigan: http://stackoverflow.com/questions/1504717/why-does-comparing-strings-in-python-using-either-or-is-sometimes-produce

msmbuilder/featurizer/featurizer.py
```diff
@@ -964,6 +967,106 @@ def describe_features(self, traj):
...
+            'closest-heavy' : distance is the closest distance between
+                any two non-hydrogen atoms in the residues
+    ignore_nonprotein : bool
+        When using `contact==all`, don't compute contacts between
+        "residues" which are not protein (i.e. do not contain an alpha
+        carbon).
+    center : float, default=0.8
+        Determines the midpoint of the sigmoid, x_0
+    steepness : float, default=2
+        Determines the steepness of the logistic curve, k
```
mpharrigan: These parameters have units I think. Might want to mention them here

mpharrigan: `nm` and `1/nm`, resp

mpharrigan:
Test failures seem relevant

msmbuilder/featurizer/featurizer.py
cxhernandez: I think steepness should be 20 `1/nm`

msmbuilder/commands/featurizer.py
brookehus: @mpharrigan the real lesson here is that i shouldn't copy and paste code

msmbuilder/featurizer/featurizer.py
```diff
@@ -964,6 +967,110 @@ def describe_features(self, traj):
...
+        distances larger than CENTER will return values smaller than 0.5.
+    steepness : float, default=20
+        Determines the steepness of the logistic curve, [1/nm]. Small 
+        and large distances will approach ouput values of 1 and 0,
+        respectively, more quickly.
+    """
+
+
+    def __init__(self, contacts='all', scheme='closest-heavy', ignore_nonprotein=True,
+                                                            center=0.8, steepness=2):
```
cxhernandez: change the default value of steepness

msmbuilder/tests/test_contactfeaturizers.py
```diff
@@ -0,0 +1,85 @@
...
+
+
+def test_binaries():
+    dataset = fetch_fs_peptide()
+    trajectories = dataset["trajectories"]
+    binarycontactfeaturizer = BinaryContactFeaturizer()
+    binaries = binarycontactfeaturizer.transform([trajectories[0]])
+
+    assert binaries[0].shape[1] == 171
+    assert sum(sum(binaries[0])) <= binaries[0].shape[0]*binaries[0].shape[1]
```
cxhernandez: You could just use `np.sum` once, but what might be more appropriate is to check that the `dtype` of this numpy object is `bool`?

msmbuilder/tests/test_contactfeaturizers.py
```diff
@@ -0,0 +1,85 @@
...
+
+
+def test_binaries_inf_cutoff():
```
cxhernandez: Could be worth adding a `test_binaries_zero_cutoff` as well

msmbuilder/tests/test_contactfeaturizers.py
```diff
@@ -0,0 +1,85 @@
...
+    steepness = 10*np.random.randn()
+    center = np.random.randn()
+    binarycontactfeaturizer = BinaryContactFeaturizer(cutoff=center)
+    binaries = binarycontactfeaturizer.transform([trajectories[0]])
+    logisticcontactfeaturizer = LogisticContactFeaturizer(center=center, steepness=steepness)
+    logistics = logisticcontactfeaturizer.transform([trajectories[0]])
+
+    # This checks that no distances that are larger than the center are logistically
+    # transformed such that they are less than 1/2
+    assert sum(sum((logistics[0]*binaries[0] < 0.5)*(logistics[0]*binaries[0] > 0.0))) == 0
```
cxhernandez:
```python
assert all((logistics[0].ravel() >= 0.5) - (binaries[0].ravel() == True) == 0)
```

msmbuilder/tests/test_contactfeaturizers.py
```diff
@@ -0,0 +1,85 @@
...
+    test = np.log(1.0/y-1)/steepness + center
+
+    np.testing.assert_almost_equal(x, test, decimal=2)
+
+
+def test_binary_to_logistics():
+    dataset = fetch_fs_peptide()
+    trajectories = dataset["trajectories"]
+    steepness = 10*np.random.randn()
+    center = np.random.randn()
```
cxhernandez: This test is throwing an error on travis-ci. My guess is that this line and the one before it are spitting out negative values.

mpharrigan: better yet, add something to the init method to assert that these are positive and give the user a helpful error message if not

msmbuilder/cluster/kcenters.py
```diff
@@ -77,6 +77,9 @@ def __init__(self, n_clusters=8, metric='euclidean', random_state=None):
         self.random_state = random_state
 
     def fit(self, X, y=None):
+        if isinstance(X, np.ndarray):
+            if X.dtype != 'float32' and X.dtype != 'float64':
```
cxhernandez: Not sure the check is really necessary. I would just go ahead and cast it:

```python
X = X.astype(float)
```

cxhernandez:
aside from my really minor comment above, this lgtm

brookehus:
I tried that, but some tests failed. I think libdistance requires that when there are multiple inputs (X and y) that they be the same type of float (either 32 or 64).

mpharrigan:
Maybe the check would be more idiomatic as
```python
if not (dype==float32 or dtype==float64)
```

But this looks good. Update changelog!

brookehus:
yuck why did the changelog get weird here??

mpharrigan:
I bet there's not supposed to be a space before the `-`s. I'll fix that in a follow-up pr

mpharrigan:
I restarted the travis failure, which is one I haven't seen before and hopefully a one-off thing. Windows tests don't even build. I'll have to boot into windows one of these days and see if I can fix things

mpharrigan:
The py35 tests keep getting killed?

Also these new tests take over 20 seconds. I'd like to stop adding new tests that are "red" for time

mpharrigan:
Actually they're like 6 seconds on the other travis runs. Why is it slower on other travis?

cxhernandez:
bump @mpharrigan

mpharrigan:
I'll restart the tests but my issues still stand

brookehus:
What are your issues specifically? I can't tell from your comments above what I need to do.

mpharrigan:
The tests fail and seem to hang in the newly added tests/code

brookehus:
I have no idea about the travis error message for the third build, `No output has been received in the last 10 minutes, this potentially indicates a stalled build or something wrong with the build itself.`, and also have no idea why `test_binary_to_logistics` takes SO long. It doesn't take that long on my laptop. Do you think I should just get rid of it?

mpharrigan:
Let's not delete it just yet. It might be indicative of a real problem.

brookehus:
Okay, well given that it runs fine and in <10s on my machine, how do you suggest gaining further insight?

cxhernandez:
`msmbuilder.tests.test_contactfeaturizers.test_binary_to_logistics ... (xdrfile error) Undocumented error 3ERROR`

could this be something weird with mdtraj then?

rmcgibbo:
Is the XTC file corrupted? That's the kind of error I would expect mdtraj to throw in that situation.

cxhernandez:
Those same XTCs working for the other featurizer tests though, right?

mpharrigan:
@brookehus Do you want to try rebasing this and letting the tests re-run? Maybe it will magically work and we can get this in to v3.5

cxhernandez:
Still something weird going on in py3.4... but doesn't seem to be a result of this PR

mpharrigan:
Wanna rebase now the #830 is merged?

cxhernandez:
This looks good now!

dr-nate:
Yayyy!

brookehus:
tbt to the never-ending #798 thread

Issue #797, opened 2016-04-10
Add Feature Selection Algorithms


cxhernandez:
We can start by:

- [ ] porting from [sklearn](http://scikit-learn.org/stable/modules/feature_selection.html)
- [ ] adding a few from [this resource](http://featureselection.asu.edu/algorithms.php) (esp the decision tree-based approaches)
- [ ] implementing [Gradient Boosted Feature Selection (GBFS)](http://alicezheng.org/papers/gbfs.pdf)

cc @msultan

msultan:
i wonder if we need to push all the features through a Normalizer before we can use some of these algorithms. We should be careful about that. The variance threshold based metrics in particular might be troublesome.  

a few technical aspects

1). It would be nice if it could accept a bunch of featurizer objects and return a folder of featurized trajs
2). Some kind of feature descriptors 
3). Our wrapper should hopefully figure out on the fly whether the fully featurized set can fit in memory and if not it should stride appropriately to fit and transform everything later

cxhernandez:
> i wonder if we need to push all the features through a Normalizer before we can use some of these algorithms. We should be careful about that. The variance threshold based metrics in particular might be troublesome.

It appears you're [right](http://stats.stackexchange.com/questions/105478/feature-normalization-standardization-before-or-after-feature-selection)!

> It would be nice if it could accept a bunch of featurizer objects and return a folder of featurized trajs
That's outside the scope of this issue, I think.

> Some kind of feature descriptors 
Agree. We should be able to pass info from `describe_features`.

> Our wrapper should hopefully figure out on the fly whether the fully featurized set can fit in memory and if not it should stride appropriately to fit and transform everything later

I think this is possible, and is done in NumPy at least (but I'm having trouble finding the check rn). Maybe @mpharrigan knows?

PR #799, opened 2016-04-11
port sklearn feature_selection

cxhernandez:
 - [x] Add tests
 - [x] Update changelog

[Describe changes here]

re: https://github.com/msmbuilder/msmbuilder/issues/797

Expanding msmbuilder pipeline functionality by porting feature selection algorithms from sklearn.


cxhernandez:
I'm noticing that [quite a few of these classes](http://scikit-learn.org/stable/modules/classes.html#module-sklearn.feature_selection) require some sort of response variable for either classification or regression, which doesn't really seem applicable to our problem scope..... I'll try to exclude these (e.g. `chi2`)

Some of them also require a scoring function, which I can only imagine in this case could be `tICA.score` or `PCA.score`?

mpharrigan:
Yeah let's be careful about only exposing things people might actually find useful!

cxhernandez:
@mpharrigan: Would it be okay to remove `msmbuilder.featurizer.FeatureUnion`?

mpharrigan:
Sounds good to me

cxhernandez:
Also, some of these, like `FeatureSelector` and `FeatureUnion`, will take lists of trajectories while the rest can only handle lists of `ndarray` objects.... Not sure if maybe we want to think about moving `FeatureSelector` and `FeatureUnion` to `msmbuilder.featurizer` (which perhaps we should rename to `feature_extraction` to be [consistent with scipy](http://scikit-learn.org/stable/modules/feature_extraction.html)?)

cxhernandez:
okay, I think this is pretty much ready for review

docs/changelog.rst
```diff
@@ -11,10 +11,21 @@ This is the current development version of MSMBuilder
 API Changes
 ~~~~~~~~~~~
 
+- ``msmbuilder.featurizer.FeatureUnion`` is now deprecated. Please used
```
mpharrigan: used -> use

msmbuilder/feature_selection/base.py
```diff
@@ -0,0 +1,119 @@
+# Author: Carlos Xavier Hernandez <cxh@stanford.edu>
```
msultan: is this code significantly different than the cluster version? Why not just inherent from that directly instead of copying it over.

cxhernandez: The docstrings should be different at least.

msultan: maybe we should have a timeseries to array concatenator and everything inherits from it. The descriptor can be different but then use the super method to initiate the base class.

cxhernandez: Is this better?

msultan: we can move the clusterer thing to a super class later on.

msmbuilder/feature_selection/featureselector.py
```diff
@@ -0,0 +1,60 @@
+# Author: Carlos Xavier Hernandez <cxh@stanford.edu>
+# Contributors:
+# Copyright (c) 2016, Stanford University and the Authors
+# All rights reserved.
+import numpy as np
+
+from .base import MultiSequenceFeatureSelectionMixin
+
+
+class FeatureSelector(MultiSequenceFeatureSelectionMixin):
```
msultan: We should add a warning that you need to normalize the dataset afterwards. Mixing and matching featurizers with differing means and variances is gonna cause a lot of issues later on. Alternate option is to add a sklearn normalizer as a attribute to the object and it get auto called when fitting.

msultan:
thanks for doing this @cxhernandez

msmbuilder/feature_selection/featureselector.py
cxhernandez: I'll make a note in the docstring, but I think a warning might be a tad much. @mpharrigan thoughts?

msmbuilder/feature_selection/featureselector.py
```diff
@@ -14,6 +14,9 @@ class FeatureSelector(MultiSequenceFeatureSelectionMixin):
     concatenates the results. This is useful to combine several feature
     extraction mechanisms into a single transformer.
 
+    Note: Users should consider using `msmbuilder.preprocessing.StandardScaler`
```
msultan: maybe i am being paranoid but a stronger warning is probably warranted. 
```
WARNING: Users are highly encouraged to use msmbuilder.preprocessing.StandardScaler to normalize the outputs ```

mpharrigan:
lgtm

cxhernandez:
@mpharrigan: can this one be merged as well?

PR #800, opened 2016-04-11
Update markov_appreciation.py

mpharrigan:


brookehus:
cc: @msultan :P

msultan:
This is what happens when we dont appreciate Andrei Markov enough.

PR #808, opened 2016-05-10
[WIP] Add SymmetryRMSDFeaturizer

mpharrigan:
 - [ ] Add tests
 - [ ] Update changelog

For homo-multi-mers, there's rotational or permutational symmetry. This featurizer takes the minimum distance from a defined set of atom index permutations.

### DIFF
--- a/msmbuilder/featurizer/__init__.py
+++ b/msmbuilder/featurizer/__init__.py
@@ -1,5 +1,7 @@
 from __future__ import absolute_import
 
+from .base import Featurizer, featurize_all
 from .feature_union import FeatureUnion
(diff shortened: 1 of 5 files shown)


mpharrigan:
I also want to start splitting up `featurizer.py`--which is huge--into files grouped by theme

brookehus:
Maybe we should iron out #798 before we split up `featurizer.py`

mpharrigan:
This just splits out the rmsd ones. We can do it bit by bit

mpharrigan:
Realized I'm going to have to do this in libdistance so it works with clustering

PR #812, opened 2016-05-13
describe features for a list of featurizers

msultan:
 - [x] Add tests
 - [ ] Update changelog

Added ability to describe features for the featureselector module. Works well for some of my stuff.


cxhernandez:
+1

mpharrigan:
Wanna add a simple test?

msultan:
added one, it was surprisingly annoying to compare the equivalence of two pandas data frames where the types were changing along

msultan:
one thing that i havent figured out is how to handle featurizers which dont have ```describe_features```. right now it will raise an error when we call it but a better option might be to add N/A to all the columns or something. 
a third option might be to add that as the default value for all featurizer objects by changing the base class featurizer class.

cxhernandez:
+1 for returning `None` for missing feature descriptions

mpharrigan:
What's the status of this?

msultan:
missing the N/A for featurizers that dont have the describe_features attribute. i think i will have to change the featurizer object a bit for that to work properly. gimme a few hours today.

PR #816, opened 2016-05-20
Override sklearn's obnoxious DeprecationWarning

mpharrigan:
screenshot from 2016-05-20 13 32 31


mpharrigan:
cc #815

msultan:
i need to learn that filter. the warnings been driving me insane. well done!

mpharrigan:
The problem is the sklearn people explicitly put in an `always` filter for their deprecation warnings and don't bother to fix the deprecation warnings. Since the warnings filter list is global, any time you import sklearn it will overwrite whatever filters you have set up (python default is to not show deprecationwarning)

The correct thing to do is only enable deprecation warnings for developers or while running tests

PR #818, opened 2016-05-25
add low-pass filtering in preprocessing

cxhernandez:
 - [x] Add tests
 - [x] Update changelog

Adding Butterworth and EWMA filters for preprocessing


msmbuilder/preprocessing/timeseries.py
```diff
@@ -0,0 +1,61 @@
...
+            order filters cutoff more quickly, but have worse numerical
+            properties.
+
+        References
+        ----------
+        .. [1] "FiltFilt". Scipy Cookbook. SciPy. <http://www.scipy.org/Cookbook/FiltFilt>.
+        """
+
+
+        if width < 2.0 or not isinstance(width, int):
```
msultan: should this be < 1.0 ?

cxhernandez: > "Must be an integer greater than one."

I guess it could be 1, but then you wouldn't really be smoothing....

msultan: nm, i read the logic wrong the first time. how about adding a default of 2 or something reasonable for this?

cxhernandez:
this works :)

Now for a debate on whether this has any merit in statistical analysis....

rmcgibbo:
I believe Matt is aware of my view.

cxhernandez:
up to you @mpharrigan

mpharrigan:
This messed up the changelog

Issue #781, opened 2016-03-29
Write GMRQ docs


mpharrigan:
The current version is a little sparse

http://msmbuilder.org/3.4.0/gmrq.html

cxhernandez:
"a little"

PR #821, opened 2016-06-07
GMRQ docs

cxhernandez:
Here's a first crack at this. Let me know what you think, and if I should include Osprey in here.

### DIFF
--- a/docs/gmrq.rst
+++ b/docs/gmrq.rst
@@ -1,8 +1,33 @@
 .. _gmrq:
 
-Model Selection (GMRQ)
+Model Selection using GMRQ
 ======================
 
-GMRQ uses a variational principle to help you select model parameters.
+The generalized matrix Rayleigh quotient (GMRQ) is a specific application of
+the variational principle (adapted from `quantum mechanics <https://en.wikipedia.org/wiki/Variational_method_(quantum_mechanics)>_`)
+for Markov state models and a useful tool for model parameter selection.
 
-.. todo: write
+The variational principle yields a rigorous way of comparing two different
+Markov models for the same underlying stochastic process when using different
+state decompositions. Even under the assumption that you have access to
+infinite sampling, there is still some error associated with approximating the
+true continuous eigenfunctions of your modeled process with the indicator
+functions, as is the case with Markov state models. If we interpret the
+variational theorem as the measure of the quality of this approximation, the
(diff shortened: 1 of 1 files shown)


docs/gmrq.rst
```diff
@@ -1,8 +1,33 @@
 .. _gmrq:
 
-Model Selection (GMRQ)
+Model Selection using GMRQ
```
mpharrigan: Check out the sidebar on msmbuilder.org. I like the idea of each page title being of the form
> High level description (technical acronym)

We tell people a lot about "GMRQ"; It should be immediately apparent where they can get a description of what this mysterious 4LA is

Also, if you want to change the title, adjust the underbar to be the same width

docs/gmrq.rst
```diff
@@ -1,8 +1,33 @@
...
 ======================
 
-GMRQ uses a variational principle to help you select model parameters.
+The generalized matrix Rayleigh quotient (GMRQ) is a specific application of
+the variational principle (adapted from `quantum mechanics <https://en.wikipedia.org/wiki/Variational_method_(quantum_mechanics)>_`)
```
mpharrigan: Is this valid syntax for links in rst?

mpharrigan:
Can we include api reference here?

Maybe we can add the gmrq vs. states plot from @rmcgibbo

docs/gmrq.rst
cxhernandez: It's what's used for the citation link.... Is this not right?

docs/gmrq.rst
cxhernandez: I guess. I just don't think GMRQ is the end-all-be-all for this. Either this section is about GMRQ or the broader topic of model selection, which can be expanded as more methods are developed.

cxhernandez:
note to self: consolidate CT-MSM section with MSM, get rid of `GMRQ` in sidebar, curate the ordering in sidebar.

docs/gmrq.rst
mpharrigan: The github render shows it weird. Maybe theres not supposed to be a space before `<`?

mpharrigan:
fixes #781

mpharrigan:
I'm going to work off of this and try to add an api reference section to this text

PR #830, opened 2016-06-21
use msmb_data

cxhernandez:
 - [x] Add tests
 - [x] Update changelog

Stop using `msmbuilder.example_datasets` (i.e. figshare) for testing

### DIFF
--- a/devtools/conda-recipe/meta.yaml
+++ b/devtools/conda-recipe/meta.yaml
@@ -42,6 +42,7 @@ test:
     - statsmodels
     - hmmlearn
     - cvxpy # [not win]
+    - msmb_data
   imports:
     - msmbuilder
   commands:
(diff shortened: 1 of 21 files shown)


cxhernandez:
@mpharrigan should we just get rid of `example_datasets` all together?

mpharrigan:
I'd imagine there are a bunch of scripts floating around in the wild that use `example_datasets`. We shouldn't break that API for no reason. Ideally, I'd like this sort of mechanism

```python
try:
  import msmb_data
  get_function = msmb_data.get()
except ImportError:
  warnings.warn("You should install msmb_data", DeprecationWarning)
  get_function = figshare.get()

def fetch_fs_peptide():
  traj = get_function()
  logic_goes_here(traj)
  return traj
```

cxhernandez:
Okay, it looks good on Travis... Not sure what's going on with Appveyor— it keeps trying to download from figshare. @mpharrigan: could you try building locally on your windows box?

Issue #832, opened 2016-06-22
Non-ASCII character in ktica.py


brookehus:
`from msmbuilder.decomposition import tICA` gives:

```
  File "/Users/brooke/source/msmbuilder/msmbuilder/decomposition/ktica.py", line 103
    .. [1] Schwantes, Christian R., and Vijay S. Pande. J. Chem Theory Comput. 11.2 (2015): 600–608.
                                                                                                    ^
SyntaxError: Non-ASCII character '\xe2' in file /Users/brooke/source/msmbuilder/msmbuilder/decomposition/ktica.py on line 104, but no encoding declared; see http://python.org/dev/peps/pep-0263/ for details
```

brookehus:
```
[brooke:~/source/msmbuilder (master)]$ conda list --export
# This file may be used to create an environment using:
# $ conda create --name <env> --file <this file>
# platform: osx-64
_license=1.1=py27_0
abstract-rendering=0.5.1=np19py27_0
alabaster=0.7.3=py27_0
anaconda=2.3.0=np19py27_0
appscript=1.0.1=py27_0
argcomplete=0.8.9=py27_0
astropy=1.0.3=np19py27_0
babel=1.3=py27_0
bcolz=0.9.0=np19py27_0
beautiful-soup=4.3.2=py27_0
binstar=0.11.0=py27_0
bitarray=0.8.1=py27_0
blaze-core=0.8.0=np19py27_0
blz=0.6.2=np19py27_1
bokeh=0.9.0=np19py27_0
boto=2.38.0=py27_0
bottleneck=1.0.0=np19py27_0
cdecimal=2.3=py27_0
certifi=14.05.14=py27_0
cffi=1.1.0=py27_0
clyent=0.3.4=py27_0
colorama=0.3.3=py27_0
configobj=5.0.6=py27_0
cryptography=0.9.1=py27_0
curl=7.43.0=0
cython=0.22.1=py27_0
cytoolz=0.7.3=py27_0
datashape=0.4.5=np19py27_0
docutils=0.12=py27_0
dynd-python=0.6.5=np19py27_0
enum34=1.0.4=py27_0
fastcache=1.0.2=py27_0
flask=0.10.1=py27_1
freetype=2.5.2=2
funcsigs=0.4=py27_0
gevent=1.0.1=py27_0
gevent-websocket=0.9.3=py27_0
greenlet=0.4.7=py27_0
grin=1.2.1=py27_1
[...]

PR #833, opened 2016-06-22
No unicode in source files

mpharrigan:
Fixes gh-832

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -20,6 +20,8 @@ New Features
 Improvements
 ~~~~~~~~~~~~
 
+- Remove a unicode character from ``ktica.py`` (gh-833)
+
 
 
 v3.5 (June 14, 2016)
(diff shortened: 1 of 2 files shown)


mpharrigan:
Do we know the scope of this problem? Do we have to release 3.5.1 with this fix backported?

mpharrigan:
@cxhernandez asks why this didn't cause failures on travis

On my machine it doesn't cause a problem either on py2 or py3 (hypothesis 1: this may be less of a problem on py3)

hypothesis 2: it depends on the environment variable `$LANG`

```bash
$ echo $LANG
en_US.UTF-8
```

hypothesis 3: breaks on mac

mpharrigan:
hypothesis 4: the encoding is set in some metadata. It got flagged to UTF-8 when someone put that dash in but that doesn't persist in git

```bash
$ file msmbuilder/decomposition/ktica.py 
msmbuilder/decomposition/ktica.py: Python script, UTF-8 Unicode text executable
$ file msmbuilder/decomposition/tica.py 
msmbuilder/decomposition/tica.py: Python script, ASCII text executable
```

PR #838, opened 2016-06-24
Fix tpt

cxhernandez:
 - [x] Add tests
 - [x] Update changelog

add logic to handle MSMs with `msm.all_populations_`

### DIFF
--- a/msmbuilder/tpt/committor.py
+++ b/msmbuilder/tpt/committor.py
@@ -79,7 +79,7 @@ def committors(sources, sinks, msm):
         commits = np.zeros(msm.all_transmats_.shape[:2])
         for i, tprob in enumerate(msm.all_transmats_):
             commits[i, :] = _committors(sources, sinks, tprob)
-        return commits.mean(axis=0)
+        return np.median(commits, axis=0)
 
     return _committors(sources, sinks, msm.transmat_)
 
@@ -142,7 +142,7 @@ def conditional_committors(source, sink, waypoint, msm):
         for i, tprob in enumerate(msm.all_transmats_):
             cond_committors[i, :] = _conditional_committors(source, sink,
                                                             waypoint, tprob)
-        return cond_committors.mean(axis=0)
+        return np.median(cond_committors, axis=0)
 
     return _conditional_committors(source, sink, waypoint, msm.transmat_)
 
--- a/msmbuilder/tpt/committor.py
+++ b/msmbuilder/tpt/committor.py
@@ -77,7 +77,10 @@ def committors(sources, sinks, msm):
     sinks = np.array(sinks, dtype=int).reshape((-1, 1))
 
     n_states = msm.n_states_
-    tprob = msm.transmat_
+    if hasattr(msm, 'all_transmats_'):
+        tprob = msm.all_transmats_.mean(0)
+    else:
+        tprob = msm.transmat_
 
     # construct the committor problem
     lhs = np.eye(n_states) - tprob
@@ -149,8 +152,11 @@ def conditional_committors(source, sink, waypoint, msm):
     .. [1] Dickson & Brooks (2012), J. Chem. Theory Comput., 8, 3044-3052.
     """
 
-    tprob = msm.transmat_
     n_states = msm.n_states_
+    if hasattr(msm, 'all_transmats_'):
+        tprob = msm.all_transmats_.mean(0)
+    else:
+        tprob = msm.transmat_
 
     # typecheck
     for data in [source, sink, waypoint]:
--- a/msmbuilder/tpt/committor.py
+++ b/msmbuilder/tpt/committor.py
@@ -78,7 +78,7 @@ def committors(sources, sinks, msm):
 
     n_states = msm.n_states_
     if hasattr(msm, 'all_transmats_'):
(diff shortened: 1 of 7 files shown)


msultan:
i am not sure this is the right way to do this. the mean of all the transition matrices in not necessarily a valid transition matrix. imo, it would be better to iterate over all the matrices and then return the mean of the transition paths.

cxhernandez:
What's the mean of a transition path....

msultan:
yea fair point. btw does this work if you have ergodic trimming in some of the matrices?

cxhernandez:
I don't really know much about this code, but I don't know why it wouldn't. It might not report the correct state mappings though....

msultan:
yea if there is trimming this is gonna give completely weird results.

cxhernandez:
Alright, I can add a fix for that— shouldn't be too bad.

msultan:
hmm, how about returning lists of tpts +fluxes for each of the matrices and letting the modeller decide what to with them?  
also, if a state is thrown out in model but not in another i am not even sure what tpt should even do in that context.

cxhernandez:
> if a state is thrown out in one model 

That's only an issue in bootstrapped msms, right? I'm not going to worry about that now....

msultan:
>That's only an issue in bootstrapped msms, right?

yea, but the all_tmats_ attribute doesnt map the transition matrix back to the mle msm labels. its just all the data.

msmbuilder/tpt/committor.py
```diff
@@ -77,7 +77,10 @@ def committors(sources, sinks, msm):
     sinks = np.array(sinks, dtype=int).reshape((-1, 1))
 
     n_states = msm.n_states_
-    tprob = msm.transmat_
+    if hasattr(msm, 'all_transmats_'):
+        tprob = msm.all_transmats_.mean(0)
```
mpharrigan: is this `axis=0`? Can you add the keyword so readers of the code won't be confused

msultan:
oh this is for bayesian msms only? i used the same name for the bootstrap version and was thinking of that. the bayesian ones shouldnt have the mapping problem. 

i think adding an explicit isinstance(msm, bayesianmsm) is necessary. i am still not sure if the mean of the tmat is the right statistic to calculate.

cxhernandez:
Yeah I had bayesian msms in mind. IMO we should fix bootstrap msm so it behaves more like bayesian msm. It's not good to keep having to add checks like this...

msultan:
i thought about that but without having to assume underlying structure on the model, its not easy. its definitely doable not by adding some prior counts to each msm.

cxhernandez:
How about something like this:

```python

def func(source, sink, msm_copy):

    if hasattr(msm, 'all_transmats_'):
        output = np.zeros_like(msm.all_transmats_)
        for i, tprob in enumerate(msm.all_transmats_):
            msm_copy = copy.copy(msm)
            msm_copy.transmat_ = tprob
            del(msm_copy.__dict__["all_transmats_"])
            output[i, :, :] = func(source, sink, msm_copy)
        return output
    else:
        tprob = msm.transmat_

    # execute code normally

```

So it returns an array of whatever it's measuring (e.g. committor values, fluxes) with an entry for each `transmat_`. You or @mpharrigan might find this hacky though....

cxhernandez:
Or we could have private functions that operate on the `transmat_` directly and just have the public functions figure out the logic...

cxhernandez:
@msmbuilder/developers: This should be ready for review whenever you get a chance

devtools/conda-recipe/meta.yaml
```diff
@@ -47,7 +47,7 @@ test:
     - msmbuilder
   commands:
     - msmb -h
-    - nosetests msmbuilder -v --with-timer
+    - nosetests msmbuilder -v --with-timer --timer-top-n 5
```
mpharrigan: why is this showing up?

msmbuilder/decomposition/ktica.py
```diff
@@ -105,11 +105,7 @@ class KernelTICA(tICA):
     def __init__(self, n_components=None, lag_time=1, shrinkage=None,
                  kinetic_mapping=False, kernel='rbf', degree=3, gamma=None,
                  coef0=1., stride=1, landmarks=None, random_state=None,
-<<<<<<< HEAD
                  kernel_params=None):
-=======
```
mpharrigan: why is this showing up?

mpharrigan:
Can you merge master? There are weird things showing up in the diff

msultan:
@cxhernandez as far as i can tell you are now gonna report the mean of the mftps, hub scores, commitors values for each state/path across all the transition matrices. Here are my two main questions. 

1. If the top path changes across the transition models, is it fair to compute the mean of the mfpts then? 

2. Based upon the convos we have had about robust statistics, is mean the right statistic to report? Perhaps the median value would take care of outliers.

cxhernandez:
> If the top path changes across the transition models, is it fair to compute the mean of the mfpts then?

It's still a valid mean IMO. The whole point is that you're trying to get a gauge of the average over many similar models.

> Based upon the convos we have had about robust statistics, is mean the right statistic to report? Perhaps the median value would take care of outliers.

This is a good point. It might be better to report the median.

Issue #849, opened 2016-07-06
Implement Median Absolute Deviation (MAD) scaling


cxhernandez:
see https://en.wikipedia.org/wiki/Robust_measures_of_scale#IQR_and_MAD

We can make this an online algorithm fairly easily as shown in this [post](http://stackoverflow.com/questions/1058813/on-line-iterator-algorithms-for-estimating-statistical-median-mode-skewnes)

PR #850, opened 2016-07-06
MADScaler

cxhernandez:
 - [x] Add tests
 - [x] Update changelog

re: https://github.com/msmbuilder/msmbuilder/issues/849

### DIFF
--- a/msmbuilder/preprocessing/_mad.pyx
+++ b/msmbuilder/preprocessing/_mad.pyx
@@ -1,0 +1,43 @@
+#!python
+#cython: language_level=3, wraparound=False, infer_types=True
+
+# Author: Carlos Xavier Hernandez <cxh@stanford.edu>
+# Contributors:
+# Copyright (c) 2016, Stanford University and the Authors
+# All rights reserved.
+
+cdef extern from "math.h":
+    double abs(double m)
+
+cimport cython
+
+
+cdef inline double sign(double x):
+  if x > 0.0:
+    return 1.
+  elif x < 0.0:
+    return -1.
+  return 0.0
+
+
+@cython.boundscheck(False)
+def mad_estimate(double[:, ::1] arr,
+                 double[::1] mad,
+                 double[::1] med,
+                 double eta):
+    cdef unsigned int i
+    cdef unsigned int j
+    cdef double dev1, dev2
+
+    cdef unsigned int n_rows = len(arr)
+    cdef unsigned int n_cols = len(mad)
+
+
+    for i in range(n_rows):
+        for j in range(n_cols):
+          dev1 = arr[i, j] - med[j]
+          med[j] += eta * sign(dev1)
+          dev2 = abs(arr[i, j] - med[j]) - mad[j]
+          mad[j] += eta * sign(dev2)
+
+    return mad, med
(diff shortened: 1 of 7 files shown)


cxhernandez:
The online algorithm (implemented in cython) is still 2 order of magnitude slowdown than a naive numpy implementation:

```python
In [1]: def mad_naive(X):
    Y = np.concatenate(X)
    med = np.median(Y, axis=0)
    mad = np.median(np.abs(Y - med), axis=0)
    return mad, med
   ....:

In [2]: from msmbuilder.preprocessing import MADScaler; mad = MADScaler()

In [23]: %timeit mad.fit(X)
1 loop, best of 3: 814 ms per loop

In [24]: %timeit mad_naive(X)
100 loops, best of 3: 19.1 ms per loop
```

Not sure if the memory savings are entirely worth it....

mpharrigan:
1. Worried about adding more cython (maintenence) esp. if it's slower
2. use `cython -a` to find hotspots
3. Your for loop is probably doing python-y things and ruining your code. Use `cdef int i; for i in range()`
4. use cython's typed memoryviews instead of the verbose numpy syntax

mpharrigan:
5. your use of high-level numpy function within the loop is certainly doing python-y things and ruining your code

cxhernandez:
Thanks for the recs! That got it down to about the same speed as the naive implementation (+/- a few hundred microseconds). 

> Worried about adding more cython (maintenence) esp. if it's slower

It's a pretty small file. I'll do some tests to see if this works better than the `RobustScaler`. Another option is to see if the `scipy` folks are interested and then we won't have to worry about it.

mpharrigan:
> Another option is to see if the scipy folks are interested and then we won't have to worry about it.

You should probably do this regardless :)

msmbuilder/preprocessing/_mad.pyx
```diff
@@ -0,0 +1,46 @@
...
+@cython.infer_types(True)
+def mad_estimate(ndarray[np.float64_t, ndim=2] arr,
+                 ndarray[np.float64_t, ndim=1] mad,
+                 ndarray[np.float64_t, ndim=1] med,
+                 double eta):
+    cdef int i
+    cdef int n_rows = arr.shape[0]
+    cdef int n_cols = arr.shape[1]
+    cdef double[:, ::1] arr2 = arr
+    cdef double[::1] mad2 = mad
```
mpharrigan: Just change the type in the function signature. It works I promise!

mpharrigan:
cdef `dev1` and `dev2`

You should be able to get it so the loop in `cython -a` has no yellow at all

msmbuilder/preprocessing/_mad.pyx
cxhernandez: Like this?

```cython
def mad_estimate(double[:, ::1] arr,
                 double[::1] mad,
                 double[::1]] med,
                 double eta):
```

mpharrigan: yes

cxhernandez:
> You should probably do this regardless :)

Looking at https://github.com/scikit-learn/scikit-learn/pull/2176#issuecomment-179303371, this seems like a pretty hard task....

mpharrigan:
Yeah it seems pretty stringent. I get where they're coming from---there's value lost when giving *too* many choices. The response to the linked comment was pretty tone deaf

PR #862, opened 2016-07-22
Update examples

mpharrigan:


mpharrigan:
Live here: http://msmbuilder.org/development/examples/index.html

There are still some notebooks in `examples/advanced-notebooks` that others can adapt to be more user friendly and include here

PR #867, opened 2016-07-26
Enable msmb_data package

mpharrigan:

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,5 +1,8 @@
 environment:
 
+  global:
+    PYTHONUNBUFFERED: on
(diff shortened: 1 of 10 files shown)


mpharrigan:
Added `MininalFsPeptide` that strides by 200 (dt = 10 ns) for use in testing. Should speed things up and fix memory errors on appveyor

mpharrigan:
Also added (back) `PYTHONUNBUFFERED` so that "BUILD START" doesn't show up after the build has finished.

mpharrigan:
Lovin' the green

Issue #864, opened 2016-07-25
Slow tests


mpharrigan:
```
msmbuilder.tests.test_featurizer.test_von_mises_featurizer_2: 79.7781s
msmbuilder.tests.test_feature_descriptor.test_VonMisesFeaturizer_describe_features: 51.4629s
msmbuilder.tests.test_workflows.test_workflow_3: 39.7338s
msmbuilder.tests.test_workflows.test_workflow_2: 28.4627s
msmbuilder.tests.test_commands.test_atomindices: 23.6336s
```

The workflows are excusable. There's no reason for a featurizer test to take over a minute

cxhernandez:
Maybe we should just get rid of it?

msultan:
why would ```test_VonMisesFeaturizer_describe_features``` take a minute. that seems way too high.

cxhernandez:
Could you just compute the pdfs for the 10 random dihedrals, instead of every single one?

https://github.com/msmbuilder/msmbuilder/blob/master/msmbuilder/tests/test_featurizer.py#L97

mpharrigan:
After #867 new slowest tests:

```
msmbuilder.tests.test_workflows.test_workflow_3: 36.6312s
msmbuilder.tests.test_workflows.test_workflow_2: 28.7269s
msmbuilder.tests.test_ghmm.test_3_state: 26.2480s
msmbuilder.tests.test_ratematrix.test_doublewell: 23.2939s
msmbuilder.tests.test_commands.test_atomindices: 20.2748s
```

PR #868, opened 2016-07-26
Shorter tests

mpharrigan:
cc #864 

Re-works some of the tests by trying to make them shorter or splitting into multiples tests.

This also changes the reporting from "top 5 slowest" to "red level slow", which is defined to be > 10 seconds. We can calibrate this


cxhernandez:
side question: Should we add the appveyor badge to the README now?

PR #874, opened 2016-08-17
Ward msmb

brookehus:
 - [x] Add tests
 - [x] Update changelog

This PR adds the `ward` linkage option to `LandmarkAgglomerative` clustering. I have developed a method for predicting cluster assignments with the `ward` objective function which, to my knowledge, is new.* This implementation maintains compatibility with everything we use `fit_predict` for, e.g. cross-validation. It also remains compatible with the use of landmarks to expedite agglomerative clustering. I have added two new parameters to `AgglomerativeClustering`, `max_landmarks` and `ward_predictor`**. `max_landmarks` is intended to be useful for searching `n_clusters` space, e.g. with `osprey`, but with the constraint that the clustering is never performed using more than some number of `max_landmarks`. I plan on uploading an article about this prediction function to arXiv in the near future which will include a proof of the algorithm and a small investigation of its results; when I do so I will add the citation to the code.

\* Assignment prediction for agglomerative clustering is sort of a sketchy problem, but the standard seems to be to apply the same objective function that was used to fit assignments. This is what we already use for our other agglomerative metrics.
** `ward_predictor` is available if a different objective function is desired to predict assignments fit with `ward`; the arXiv article, in prep., addresses this.

cc: @rmcgibbo


brookehus:
The tests are failing because [here](https://github.com/brookehus/msmbuilder/blob/ward-msmb/msmbuilder/cluster/agglomerative.py#L22-L25), if we don't use fastcluster, but instead divert to scipy, it won't work.

msmbuilder/cluster/agglomerative.py
```diff
@@ -227,3 +275,4 @@ def fit_predict(self, X):
 
 class LandmarkAgglomerative(MultiSequenceClusterMixin, _LandmarkAgglomerative, BaseEstimator):
```
cxhernandez: What's the point of having `_LandmarkAgglomerative` as a class rather than just defining everything under the public class, `LandmarkAgglomerative`?

brookehus: I'm not sure- I didn't change anything from @rmcgibbo's preexisting `LandmarkAgglomerative` that wasn't necessary to accommodate the Ward stuff.

mpharrigan:
I'll review this later. From a quick scan, it looks like there are some pep8 violations

msmbuilder/cluster/agglomerative.py
```diff
@@ -197,21 +242,33 @@ def predict(self, X):
 
         dists = cdist(X, self.landmarks_, self.metric)
 
-        try:
-            pooling_func = POOLING_FUNCTIONS[self.linkage]
-        except KeyError:
-            raise ValueError('linkage=%s is not supported' % self.linkage)
+        if self.linkage == 'ward':
```
mpharrigan: I don't think this branching is necessary

msmbuilder/cluster/agglomerative.py
```diff
@@ -197,21 +242,33 @@ def predict(self, X):
...
         labels = np.zeros(len(X), dtype=int)
 
         for i in range(self.n_clusters):
             if np.any(self.landmark_labels_ == i):
-                d = pooling_func(dists[:, self.landmark_labels_ == i])
+                d = pooling_func(dists[:, self.landmark_labels_ == i],
+                                 self.cardinality_[i],
+                                 self.squared_distances_within_cluster_[i])
+                if np.any(d < 0):
+                    warnings.warn('Distance shouldn\'t be negative.')
```
mpharrigan: Use double quotes if your string contains a single quote

msmbuilder/cluster/agglomerative.py
```diff
@@ -197,21 +242,33 @@ def predict(self, X):
...
                 mask = (d < pooled_distances)
                 pooled_distances[mask] = d[mask]
                 labels[mask] = i
+            else:
+                print('No data points were assigned to cluster %c' % i)
```
mpharrigan: `.format`

msmbuilder/cluster/agglomerative.py
mpharrigan:
```python
pfunc_name = self.ward_predictor if self.linkage == 'ward' else self.linkage
try:
  pooling_func = POOLING_FUNCTIONS[pfunc_name]
except KeyError:
  raise ...
```

mpharrigan:
Nice!

msmbuilder/cluster/agglomerative.py
```diff
@@ -30,21 +27,29 @@
 
 __all__ = ['_LandmarkAgglomerative']
 
+
+def ward_pooling_function(x, cluster_cardinality, intra_cluster_sum):
+    normalization_factor = cluster_cardinality*(cluster_cardinality+1)/2
```
philipithomas: Be careful with this - in python 2, it will do integer division (i.e. dividing by 2 vs 2.0 returns an integer and a float, respectively).

brookehus: @mpharrigan do our future imports deal with this and/or should I fix?

mpharrigan: Yes, https://github.com/brookehus/msmbuilder/blob/c514a2f39ae1b375890ba4c7f77dabbe6707294d/msmbuilder/cluster/agglomerative.py#L10 should take care of it

PR #878, opened 2016-08-19
[project_templates] Save RAM and fix landmarks

mpharrigan:
Ran into the issue described in the change docstring


cxhernandez:
@mpharrigan: Is this okay to merge?

PR #879, opened 2016-08-22
test to make sure everything shows up in command line

brookehus:
 - [x] Add tests
 - [ ] Update changelog

Test was written by @mpharrigan (#737); I have tried to put in all the command line stuff.

Outstanding problems:
- Test has a problem with the following not being present in `msmb -h`, but I'm not sure if they belong there. @mpharrigan - should we edit test or do any of them belong?
1. `TrajFeatureUnion`
2. `FirstSlicer`
3. `FeatureUnion`
4. `Slicer`
5. `MultiSequenceDecompositionMixin`

- `FastICA` errors as not listed but it is listed

- Docstring for `FastICA` and `FactorAnalysis` are not good- whoever put these in can you fix? @cxhernandez?


brookehus:
Also the test takes like 2 minutes

mpharrigan:
> FastICA errors as not listed but it is listed

I think I made the regex look for camel case. You can modify it
`([A-Z][a-z])+` --> `([A-Z][A-Za-z]+)`

mpharrigan:
> Also the test takes like 2 minutes on my desktop

Let's add a `@skip` decorator to the one that actually runs `msmb -h` for each test.

mpharrigan:
> Test has a problem with the following not being present in msmb -h, but I'm not sure if they belong there. @mpharrigan - should we edit test or do any of them belong?

Should update the test

msmbuilder/tests/test_commands_exist.py
```diff
@@ -0,0 +1,125 @@
...
+            yield ma.group(1)
+        elif ' ' in line or line.endswith('.'):
+            # If the command name is too long, it forces the description
+            # to start on the next line. This is a heuristic for if this
+            # is a "description only" line
+            print("Text:", line)
+        else:
+            # See above, this is a regex for a 'command only' line.
+            # It regex tests for camelcase
+            ma = re.match(r"([A-Z][a-z])+", line)
```
mpharrigan: this looks completely wrong

mpharrigan: Maybe something like:
```([A-Z][A-Za-z]*)+```

(this wouldn't match `tICA`)

Comment (brookehus) [inline msmbuilder/tests/test_commands_exist.py]: Switching it to this doesn't appear to change anything. See above

**Comment by brookehus** [comment]
FastICA doesn't work because of the sklearn docstring. I suspect because it has a colon in it?? Do we have any precedent for editing the `__doc__` from sklearn?

**Comment by brookehus** [comment]
Current fix:

```
class FastICA(MultiSequenceDecompositionMixin, _decomposition.FastICA):
    __doc__ = ('A fast algorithm for Independent Component Analysis.\n'
               +_decomposition.FastICA.__doc__)
```

**Comment by brookehus** [comment]
Should pass now, but still very slow. Not sure how to skip stuff to make it not-slow yet.

**Comment by brookehus** [comment]
I think this is good now, albeit a bit janky.

File: msmbuilder/commands/featurizer.py [hunk 2]
```diff
@@ -199,3 +204,38 @@ def _solvent_indices_type(self, fn):
...
     def _solute_indices_type(self, fn):
         return np.loadtxt(fn, dtype=int, ndmin=1)
+
+
+class VonMisesFeaturizerCommand(FeaturizerCommand):
+    _concrete = True
+    klass = VonMisesFeaturizer
+
+
+class FunctionFeaturizerCommand(FeaturizerCommand):
```
mpharrigan: how do we plan on allowing functions to be specified via command line?

brookehus:
```
FAIL: test_msmbuilder.decomposition.FastICA_listed
Traceback (most recent call last):
  File "/Users/brooke/anaconda/lib/python2.7/site-packages/nose/case.py", line 197, in runTest
    self.test(*self.arg)
  File "/Users/brooke/source/msmbuilder/msmbuilder/tests/test_commands_exist.py", line 121, in __call__
    assert self.command in HELP_COMMANDS, err
AssertionError: FastICA was not listed in `msmb -h`
```

brookehus:
😁

msmbuilder/commands/featurizer.py
brookehus: Yeah, this needs some sort of `_def_function` or if it is unsuitable for command line then either (a) a way to classify whether functions are suitable or unsuitable from the outset or (b) more hacking of the test

msmbuilder/cluster/__init__.py
```diff
@@ -72,11 +72,5 @@ class SpectralClustering(MultiSequenceClusterMixin, cluster.SpectralClustering,
     __doc__ = _replace_labels(cluster.SpectralClustering.__doc__)
 
 
-class AgglomerativeClustering(MultiSequenceClusterMixin,
```
mpharrigan: Why is this removed?

msmbuilder/commands/featurizer.py
mpharrigan: We shouldn't include anything that can't actually be used from the command line

msmbuilder/cluster/__init__.py
brookehus: I couldn't find any place where it was actually referenced. No init or doc, e.g.

```
[brooke:~/source/msmbuilder/msmbuilder/cluster (command-line)]$ grep AgglomerativeClustering *
[brooke:~/source/msmbuilder/msmbuilder/cluster (command-line)]$ grep SpectralClustering *
__init__.py:           'GMM', 'SpectralClustering', 'KCenters', 'NDGrid',
__init__.py:class SpectralClustering(MultiSequenceClusterMixin, cluster.SpectralClustering,
__init__.py:    __doc__ = _replace_labels(cluster.SpectralClustering.__doc__)
Binary file __init__.pyc matches
```

mpharrigan: It's part of the public API. You can't just remove it

brookehus: Ok, makes sense. And it doesn't need an explicitly stated doc or anything?

brookehus: Should it be in `__all__` though?

brookehus:
So does it make sense to manually put everything that wouldn't make sense in the command line, e.g. `FunctionFeaturizer`, in the `unexposed` list [in the test](https://github.com/msmbuilder/msmbuilder/pull/879/files#diff-9f89d27f5fdaff5abe144f46949ce1a1R53), or pursue some other solution?

mpharrigan:
what was wrong with my existing list of ignores?

brookehus:
Nothing is wrong with existing list 😺

msmbuilder/commands/example_datasets.py
```diff
@@ -57,3 +57,9 @@ class MetEnkephalinDatasetCommand(DatasetCommand):
     _concrete = True
     klass = MetEnkephalin
     description = 'Download example Met-Enkephalin dataset.'
+
+
+class MinimalFsPeptideDatasetCommand(DatasetCommand):
```
mpharrigan: I don't think this will work (meaningfully) from the command line. It uses the same data as FsPeptide but loads with a stride parameter. The command line commands just dump the data, so it would be the same as FsPeptide

brookehus: Cool, so would you just remove the option?

mpharrigan: Yes

brookehus:
bump @msmbuilder/developers

Issue #884, opened 2016-09-07
verbosity option for example_datasets


cxhernandez:
Currently, if you load a dataset it'll print out:

```
loading trajectory-1.xtc...
loading trajectory-10.xtc...
loading trajectory-11.xtc...
loading trajectory-12.xtc...
loading trajectory-13.xtc...
loading trajectory-14.xtc...
....
```

this looks pretty bad in notebooks and is somewhat unnecessary for power users. It'd be nice to have the option to silence this (e.g. `verbose=False`).

cxhernandez:
done in #891

PR #891, opened 2016-09-09
Add verbose flag to _MDDataset

cxhernandez:
re: https://github.com/msmbuilder/msmbuilder/issues/884


mpharrigan:
We really should look into using python logging at some point in the future. @rmcgibbo was always against it

nbdnbdnbd:
+1 logging is great

rmcgibbo:
:)

PR #897, opened 2016-09-13
doc fix

cxhernandez:
get around msmexplorer having msmbuilder as a dependency


mpharrigan:
Would this be fixed by renaming our tested package `msmbuilder` instead of `msmbuilder-dev`

cxhernandez:
probably. Is there any reason why it's `msmbuilder-dev`?

mpharrigan:
we used to upload it. now, no reason

PR #915, opened 2016-10-04
Fix compatibility with sklearn >= 0.18

mpharrigan:
GMM was re-implemented as GaussianMixtureModel, with no deprecation cycle :(

https://github.com/scikit-learn/scikit-learn/pull/7295


mpharrigan:
Actually, they still provide an alias for `GMM`, but we scrape the docstring, hence error

mpharrigan:
I don't know why travis is failing to import numpy (cannot reproduce locally)

mpharrigan:
When "rendering" the `meta.yaml` template, conda-build is trying to scrape some data from `setup.py`. We import numpy in `setup.py` which apparently isn't available to the environment running conda build (probably the root environment) (probably why it works locally (I have numpy in my root env))

Ideally, we shouldn't need to run setup.py to "render" our `meta.yaml` file

Relevant conda-build issue thread https://github.com/conda/conda-build/issues/1429

jchodera:
> I don't know why travis is failing to import numpy (cannot reproduce locally)

We started having this problem as well: https://github.com/choderalab/yank/pull/509#issuecomment-251379910

Pinning `conda-build` to 2.0.2 seemed to fix that for us.

Lnaden:
Near as I can tell, `conda-build` 2.0.3 seems to have forgotten to require numpy, which is something they added in to require a while back, don't know what changed in 2.0.3. ¯\_(ツ)_/¯

mpharrigan:
Conda build is fixed so I restarted the tests. I don't understand the failing test, but it's probably a legit sklearn 0.18 thing. Merging this though

mpharrigan:
https://github.com/omnia-md/conda-recipes/pull/615

mpharrigan:
New 3.6.1 packages built successfully. I tested the `rc` channel (briefly). Moving to main channel. `conda update msmbuilder` should now contain these changes

PR #916, opened 2016-10-10
k-sparse tICA

maxentile:
 - [x] Add tests
 - [ ] Update changelog

Proposes a variant of sparse tICA, where the sparsity level is specified in terms of the number of nonzero entries `k` you'll tolerate in each component, rather than a regularization strength `rho` between 0 and 1. May be of particular interest to @rmcgibbo !

Under the hood, the only real difference compared with SparseTICA is that it uses [the truncated power method](http://www.jmlr.org/papers/volume14/yuan13a/yuan13a.pdf), rather than ADMM, to recover each sparse component. 

Hopefully this has a couple benefits:
- It may be a bit more intuitive to use (I don't really know what value of `rho` to choose, unless I try a few values and see how sparse the resulting solutions are, or do cross-validation, but I can probably pick `k` a priori).
- Recovering each component appears to be faster in some cases than in SparseTICA.
- It's simple to implement.

Some sanity checks I've already done:
- Comparing the pseudo-eigenvalues vs. the actual tICA eigenvalues on [the test case](https://github.com/maxentile/msmbuilder/blob/k-sparse-tica/msmbuilder/tests/test_ksparsetica.py), for various values of `k`:
image
- Checking that, in the case of `k=n_features`, this recovers the same eigenvalues as standard tICA.

Some remaining issues:
- As in SparseTICA, I don't think there's any guarantee that this recovers components in the correct order. I.e. if you ask for `n_components=10`, you may encounter an approximate eigenvector with a larger pseudoeigenvalue than if you ask for `n_components=1`. How should we handle this?
- Sometimes the pseudoeigenvalues are greater than one. Any ideas why this happens / how to handle this?
- Is there a more efficient way to handle the 1-sparse case?
  - Since there are only (at most) `n_features` 1-sparse vectors to try at each iteration, one thought is to just compute the pseudoeigenvalue of each possible choice, pick the best one, deflate, and repeat...
- I haven't done any systematic performance comparison yet. What tests would be informative?
- I haven't tried to optimize the truncated power method implementation yet.
- I haven't tried any other deflation methods yet.

### DIFF
--- a/msmbuilder/decomposition/ksparsetica.py
+++ b/msmbuilder/decomposition/ksparsetica.py
@@ -112,7 +112,7 @@ class KSparseTICA(tICA):
         Journal of Machine Learning Research. Vol. 14. 2013.
         http://www.jmlr.org/papers/volume14/yuan13a/yuan13a.pdf
     .. [2] McGibbon, R. T. and Pande, V.S. "Identification of simple reaction
-        coordinates from complex dynamics." arXiv:1602.08776 [cond-mat.state-mech]
+        coordinates from complex dynamics." arXiv:1602.08776 [cond-mat.stat-mech]
         2016. https://arxiv.org/abs/1602.08776
     .. [3] Mackey, L. "Deflation Methods for Sparse PCA." NIPS. Vol. 21. 2008.
         http://web.stanford.edu/~lmackey/papers/deflation-nips08.pdf
--- a/msmbuilder/decomposition/ksparsetica.py
+++ b/msmbuilder/decomposition/ksparsetica.py
@@ -1,0 +1,207 @@
+from __future__ import print_function, division, absolute_import
+from six import PY2
+import numpy as np
+from msmbuilder.decomposition.tica import tICA
+from msmbuilder.decomposition._speigh import scdeflate
+
+__all__ = ['KSparseTICA']
+
+
+class KSparseTICA(tICA):
+    """
+    A special case of approximate time-structure Independent Component Analysis (tICA),
+    where each approximate eigenvector is required to have k or fewer nonzero values.
+
+    This case has the nice properties that:
+    - It may be intuitive for the user to choose `k` rather than continuous parameters for regularization strength.
+    - k-sparse approximate eigenvectors can be recovered using the truncated power method, which has no additional
+        hyperparameters and typically converge quickly
+    - We recover standard tICA when k = n_features
+
+    Here's what's happening under the hood:
+    0. Turn the generalized eigenvalue problem A v = u B v into an eigenvalue problem C v = u C v
+        A = self.offset_correlation_; B = self.covariance_; C = B^{-1} A;
+
+    1. Recover the dominant* k-sparse eigenvector of C using the
+        truncated power method [1]
+
+    2. Remove the "influence" of v from C by Schur's complement deflation
+        C = (C - (C x) * (x C)) / (x C x)
+
+    3. Repeat 1,2 until the desired number of components has been recovered.
+
+    .. warning::
+        This model is currently  experimental, and may undergo significant
+        changes or bug fixes in upcoming releases.
+
+    .. warning::
+        The pseudo-eigenpairs can be recovered in the wrong order, i.e. the pseudo-eigenvalue
+        of component 5 might be greater than the pseudo-eigenvalue of component 4.
+        To recover the top-5 eigenpairs, you should really recover all of the eigenpairs, then sort them.
+
+    .. warning::
+        Sometimes the pseudoeigenvalues can exceed 1.
+        It's unclear why this happens...
+
+    .. note::
+        Haven't tested different deflation methods here yet.
+
+    Parameters
+    ----------
+    n_components : int
+        Number of sparse tICs to find.
+
+    lag_time : int
+        Time-lagged correlations are computed between X[t] and X[t+lag_time].
+
+    k : int
+        Desired sparsity of components.
+
+    shrinkage : float, default=None
+        The covariance shrinkage intensity (range 0-1).
+
+    kinetic_mapping: boolean
+        If True, scale the projection onto each eigenvector by its corresponding
+        eigenvalues, yielding the kinetic distance as described in [4].
+
+    Attributes
+    ----------
+    components_ : array-like, shape (n_components, n_features)
+        Components with maximum autocorrelation.
+
+    offset_correlation_ : array-like, shape (n_features, n_features)
+        Symmetric time-lagged correlation matrix, `C=E[(x_t)^T x_{t+lag}]`.
+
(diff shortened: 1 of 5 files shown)


mpharrigan:
Nice! I'd be curious to hear others' thoughts, but this looks good to me! Major bonus points for good documentation and plots included in pr! :)

msmbuilder/decomposition/ksparsetica.py
```diff
@@ -0,0 +1,207 @@
...
+        # recover components
+        for i in range(self.n_components):
+            x0 = self._normalize(np.random.rand(len(A)))
+            eig_vecs.append(self._truncated_power_method(BinvA, x0, k=self.k))
+            BinvA = scdeflate(BinvA, eig_vecs[-1])
+        eig_vecs = np.vstack(eig_vecs)
+
+        # compute pseudoeigenvalues
+        vs = eig_vecs.T
+        eig_vals = np.diag((vs.T.dot(A).dot(vs)).dot(np.linalg.inv(vs.T.dot(B).dot(vs))))
```
msultan: just out of curiosity, would it be correct to save a copy of the original ```BinvA``` and do 
``` np.diag(v.T.dot(BinvA).dot(v)) ```.

File: msmbuilder/decomposition/ksparsetica.py [hunk 2]
```diff
@@ -0,0 +1,207 @@
...
+        ''' given a vector x, leave its top-k absolute-value entries alone, and set the rest to 0 '''
+        not_F = np.argsort(np.abs(x))[:-k]
+        x[not_F] = 0
+        return x
+
+    def _normalize(self, x):
+        ''' given a vector x, normalize it '''
+        return x / np.linalg.norm(x)
+
+    def _truncated_power_method(self, A, x0, k, max_iter=10000, thresh=1e-8):
```
mpharrigan: meh just do that if it makes sense when implementing those things

maxentile: Whichever you both prefer is fine with me! I think it makes sense to move those three functions into _speigh.pyx , since the sparse eigenvalue solver for SparseTICA also lives there?

msmbuilder/decomposition/ksparsetica.py
```diff
@@ -0,0 +1,207 @@
...
+        '''
+
+        xts = [x0]
+        for t in range(max_iter):
+            xts.append(self._normalize(self._truncate(np.dot(A, xts[-1]), k)))
+            if np.linalg.norm(xts[-1] - xts[-2]) < thresh: break
+        return xts[-1]
+
+    def _solve(self):
+        if not self._is_dirty: return
```
maxentile: Good point! I'll fix that. I had just copied that from the SparseTICA class. Should I change this also in SparseTICA?

msmbuilder/decomposition/ksparsetica.py
maxentile: Did a quick check and these two ways of computing eigenvalues appear to give different answers on a couple test cases. I think this may be partially because A and B tend to have very high condition numbers? In any case, I figured it was better to compute these exactly the same way as the eigenvalues would be computed in GMRQ, so that I'd be sure that `kstica.score_ == kstica.score(X_train)`.

msultan: ya, I think its best to leave it as is then.

msmbuilder/decomposition/ksparsetica.py
msultan: @msmbuilder/developers any opinions on this?

maxentile:
Unsure how to address the current test failure: ["AssertionError: KSparseTICA was not listed in `msmb -h`"](https://travis-ci.org/msmbuilder/msmbuilder/jobs/167343028#L2023-L2030)

Also, looks like the test I added is more time-consuming than any other test (~80seconds, mostly because it calls the k-sparse eigensolver ~400 times), so I'll try to cut it down to a few seconds...

mpharrigan:
> Unsure how to address the current test failure: "AssertionError: KSparseTICA was not listed in msmb -h"

@brookehus can you help with this? We could merge this PR and add it to the command line in a subsequent pr if that works best for people

> Also, looks like the test I added is more time-consuming than any other test (~80seconds, mostly because it calls the k-sparse eigensolver ~400 times), so I'll try to cut it down to a few seconds...

Thanks!

brookehus:
To get this to show up in the command line,

0. Go to `msmbuilder/msmbuilder/commands`
1. In `fit_transform.py`, add `KSparseTICA` [here](https://github.com/msmbuilder/msmbuilder/blob/master/msmbuilder/commands/fit_transform.py#L25-L26)
2. At the bottom of `fit_transform.py`, write a short class ([example](https://github.com/msmbuilder/msmbuilder/blob/master/msmbuilder/commands/fit_transform.py#L183-L187))
3. In `__init__.py` (still in the `commands` subdirectory), add `KSparseTICA` to the [`.fit_transform` imports](https://github.com/msmbuilder/msmbuilder/blob/master/msmbuilder/commands/__init__.py#L24-L36)

maxentile:
Thanks, @brookehus , and sorry for my slow response!

I added these commands, but it still looks I may be missing something, since this test still isn't passing ([relevant part of Travis log](https://travis-ci.org/msmbuilder/msmbuilder/jobs/167699973#L2027)). Do you know what else I should add here to address this?

brookehus:
Hey. I pushed a change to your branch (I didn't realize this would go through automatically I imagined it would ask me to PR- sorry!). The issue was [here](https://github.com/msmbuilder/msmbuilder/pull/916/files#diff-ca17963aff9aa7a0fbbcc1b3740f40dfR11), where you have to adhere to our docstring conventions by having the first line under the class definition be something like `"""Description here` with no line break after the `"""` and where the description is online one line.

Can you add a blank line to `msmbuilder/commands/fit_transform.py` and also make sure that your additions adhere to pep8 conventions? For example, all lines should contain 80 or fewer characters. Pycharm is good at finding pep8 violations (cc: @mpharrigan).

maxentile:
Ahh, thanks for fixing this! I'll correct the remaining pep8 violations now

mpharrigan:
Thanks for the contribution @maxentile !

brookehus:
`msmbuilder/decomposition/ksparsetica.py` still has long lines in the documentation- thoughtsj @mpharrigan?

mpharrigan:
I've never been a huge fan of the 80 character limit. I do it myself, and sometimes it prevents me from jamming too many things onto one line (a good thing). What really counts is readability. I could find examples of readable lines that exceed 80 characters as well as unreadable `\`-wrapped lines that conform to the 80 character limit. [I can also find unreadable long lines and readable short lines!].
As long as it doesn't wrap in the github split diff view, I'm more-or-less fine :). Especially for outside contributors, we can be a little lenient.

PR #919, opened 2016-10-14
save all the models

msultan:
 - [x] Add tests
 - [x] Update changelog

Adds option to bootstrap msm to optionally save all models.

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -6,6 +6,11 @@ Changelog
 v3.7 (development)
 ------------------
 
+New Features
+~~~~~~~~~~~~
+
+-``BootStrapMarkovStateModel`` can now optionally save all the
+models that it generates(gh-919)
 
 v3.6 (September 15, 2016)
 -------------------------
--- a/msmbuilder/msm/validation/bootstrapmsm.py
+++ b/msmbuilder/msm/validation/bootstrapmsm.py
@@ -102,7 +102,7 @@ class BootStrapMarkovStateModel(_MappingTransformMixin):
 
         self._succesfully_fit = 0
         self._ommitted_trajs_ = None
-        self.all_models = None
+        self.all_models_ = None
         self.all_populations_ = None
         self.mapped_populations_ = None
         self.all_training_scores_ = None
@@ -164,7 +164,7 @@ class BootStrapMarkovStateModel(_MappingTransformMixin):
                     self.all_test_scores_.append(np.nan)
 
         if self.save_all_models:
-            self.all_models = all_models
+            self.all_models_ = all_models
 
     @property
     def mapped_populations_mean_(self):
--- a/msmbuilder/msm/validation/bootstrapmsm.py
+++ b/msmbuilder/msm/validation/bootstrapmsm.py
@@ -164,7 +164,7 @@ class BootStrapMarkovStateModel(_MappingTransformMixin):
                     self.all_test_scores_.append(np.nan)
 
         if self.save_all_models:
-            self.all_models = all_mdls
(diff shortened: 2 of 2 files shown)


mpharrigan:
Why does this need to be configured? Can't we just do it? Can we spell out "model"

Other things I noticed in this class:

 -  fit should return `self`
 - the msm kwargs shouldn't have a default param of `{}` because python is dumb and then this dictionary becomes suprisingly mutable. Do the normal thing of setting it to `None` and then checking `if msm_kwargs is None: msm_kwargs = {}`
 - There's a superfluous return https://github.com/msultan/msmbuilder/blob/0d1d9b35a3a4402f6d1e683a0c005a716d8f55b6/msmbuilder/msm/validation/bootstrapmsm.py#L164

mpharrigan:
Usually when we add new arguments, we put them at the end for backwards compatibility

msultan:
> Why does this need to be configured? 

Because, it might increase the object size by a lot. 

>Can't we just do it? Can we spell out "model"

Fine, but mdls is the web3.0/start-up version of this . 

> Other things I noticed in this class:
> 
> fit should return self

kk

> the msm kwargs shouldn't have a default param of {} because python is dumb and then this dictionary becomes suprisingly mutable. Do the normal thing of setting it to None and then checking if msm_kwargs is None: msm_kwargs = {}

Didnt know this,changing it now. 

> There's a superfluous return https://github.com/msultan/msmbuilder/blob/0d1d9b35a3a4402f6d1e683a0c005a716d8f55b6/msmbuilder/msm/validation/bootstrapmsm.py#L164

msultan:
No idea why py3.5 is failing and 3.4 is working somehow.

docs/changelog.rst
```diff
@@ -6,6 +6,11 @@ Changelog
 v3.7 (development)
 ------------------
 
+New Features
+~~~~~~~~~~~~
+
+-``BootStrapMarkovStateModel`` can now optionally save all the
```
cxhernandez: +1

msmbuilder/msm/validation/bootstrapmsm.py
```diff
@@ -107,7 +114,7 @@ def fit(self, sequences, y=None, pool=None):
         sequences = list_of_1d(sequences)
         self.mle_.fit(sequences, y=y)
         self._parallel_fit(sequences, pool)
-
+        return self
```
cxhernandez: ?

mpharrigan:
How big could all the models possibly be? The way it's set up now, they have to all fit in RAM simultaneously. A couple more megabytes on disk isn't going to hurt anyone (how much space are we talking here?). What about encouraging users to 
```python
msm.fit()
msm.all_models_ = None
dump(msm)
```
where space is a premium?

mpharrigan:
Should this be `msm.all_models_` (instead of `msm.all_models`)?

msultan:
for a 300 stat-space, its about 100MB for 50 bootstrap samples.

msultan:
I think if users run 1000s of bootstrap samples, this might get large.

msultan:
i changed the behavior the mapped populations so that it defaults to np.nan instead of 0s.

msultan:
@msmbuilder/developers anyone have any ideas as to why the featurizer tests are failing on p3.5 and 2.7?

mpharrigan:
it's sklearn 0.18  see #918

mpharrigan:
Is this mergeable?

msultan:
yea, tests are still failing though.

mpharrigan:
They aren't relevant to this PR

PR #920, opened 2016-10-16
FeatureSelector: Make the order in which features are added and computed deterministic.

dr-nate:
 - [ ] Add tests
 - [ ] Update changelog

This will help avoid issues where a script is run multiple different
times or on different datasets, which currently results in randomly ordered
features as output and potentially gibberish MSM results.


mpharrigan:
Oh dear :)

PR #931, opened 2016-10-24
Added error bars to MFPT

jadeshi:
- [x] Add tests
- [x] Update changelog

Added a function, generate_mfpt_dist, that allows you to generate a distribution of MFPTs accounting for errors in the transition matrix. It works by generating a Gaussian random variable for each nonzero transition probability, with standard deviation equal to the standard error (tprob - tprob ^ 2) / (total row counts) ^ 0.5. You then sample from each random variable once to get a perturbed transition matrix, compute the MFPTs for the perturbed matrix, and repeat for however many samples you want. Returns a (n_samples x n_states) 2d numpy array of MFPTs for a given set of sinks.


msultan:
hmm, @jadeshi how about combining the distribution function into the regular mfpt code and adding a new kwarg ```generate_distribtution``` or something like that. That way we can dont have to force new users to learn more function names. 

I am good with this commit otherwise though. I think the functions need to be documented though, and make it clear that the perturbation to the tmats are coming from a binomial approximation to the transition probability with the means/variance of the binomial following a gaussain based off the central limit theorem.

cxhernandez:
Maybe just call the kwarg `error`?

jadeshi:
@msultan @cxhernandez Good points, and thanks for the feedback. I incorporated the function into the original mfpts function, and added some documentation.

msmbuilder/tpt/mfpt.py
```diff
@@ -26,7 +26,45 @@
...
+        The matrix of standard deviations of the Gaussians. For transition probability (i,j), this is 
+        assumed to be the standard error of the mean of a binomial distribution with p = transition probability 
+        and number of observations equal to the summed counts in row i.
+    '''
+    output = np.vectorize(np.random.normal)(loc, scale)
+    output[np.where(output < 0)] = 0
+    return (output.transpose() / np.sum(output, axis=1)).transpose()
+
+
+def mfpts(msm, sinks=None, lag_time=1., errors='-1'):
```
cxhernandez: Could you make errors a boolean (`True`|`False`)?

msmbuilder/tpt/mfpt.py
```diff
@@ -81,7 +127,13 @@ def mfpts(msm, sinks=None, lag_time=1.):
             mfpts[i, :, :] = _mfpts(tprob, populations, sinks, lag_time)
 
         return np.median(mfpts, axis=0)
```
cxhernandez: Not sure how we want to handle this for continuous MSMs 😕

msmbuilder/tpt/mfpt.py
jadeshi: I wanted the user to be able to set how many samples to pull, so they could judge themselves how many samples it takes to give a good Gaussian MFPT distribution.

cxhernandez: So why is it a string then?

cxhernandez: For user ease, I would make `error` a kwarg set to `False` by default and then have a separate kwarg, `n_samples` (set to 30 by default), that more experienced users can fine-tune.

cxhernandez: But maybe @msmbuilder/developers have an opinion on this?

jadeshi: Yeah that's a good point. I could add another kwarg if we want to set a default value for n_samples. Also, the '-1' is just supposed to represent "False", since that's how I've seen it done in a lot of old msmbuilder 2 code, but that's probably a bit confusing. I'll change it.

jadeshi:
Anyone know why this appveyor test has been pending for several days?

cxhernandez:
I'm not sure. I'm having this issue on #933 as well

bump @mpharrigan

jadeshi:
@cxhernandez Merge?

cxhernandez:
Sorry to add another hoop to jump through, but could you add a test and update the changelog. @brookehus or @msultan can help you with this.

jadeshi:
Updated changelog and added a simple test. 
@cxhernandez Could you delete the other two PR's I made above this one? (I'm a total newb with GitHub)

cxhernandez:
The test isn't formatted quite right, but maybe @brookehus can you help you out? I'll try to be in lab this evening if need be.

brookehus:
@jadeshi - your test looks more like a demonstration of the utility. It would be awesome to include that and the plots you made as an example ipynb (python3) in our [examples directory](https://github.com/msmbuilder/msmbuilder/tree/master/examples) to demonstrate to users how one would use the mfpt error bars.

The tests are intended both to make sure your code does what you think it's doing and also to make sure other people can't break it in the future. Therefore you want to verify data structure shapes, boundary conditions, standard and weird kinds of input, etc. Please see [here](https://github.com/msmbuilder/msmbuilder/blob/master/msmbuilder/tests/test_alphaanglefeaturizer.py#L19-L26) for an example of how that would be formatted. Everything should be defined as a function `test_`+something descriptive, and should assert a truth. If that assertion fails, the test fails. Can you create this kind of test file?

jadeshi:
@brookehus Thanks a lot for the feedback. Definitely should have looked at what the format for the tests were before making one. Just uploaded an updated version and also made some very minor code fixes.

brookehus:
@jadeshi awesome, no worries! Let's see how the builds turn out

brookehus:
Awesome thanks @cxhernandez and @jadeshi

PR #940, opened 2016-11-09
Correct indentation level in describe_features

msultan:
 - [x ] Add tests
 - [ ] Update changelog

I have no idea how the dataframe was been returned when the function had extra tabs in it.


brookehus:
interesting

mpharrigan:
Is it actually tabs or is it just indented too far? I think python will accept it if it's at least consistent indentation

mpharrigan:
```python
def f(x):
    z = x**2
    return z

def g(x):
        z = x**3
        return z
    
def h(x):
    z = x**3
     return z # IndentationError
```

msultan:
i am not really sure. it just kept doing weird things on my mac and seems to have been fixed now.

Issue #935, opened 2016-11-04
implement landmark option with RMSDFeaturizer


brookehus:

msultan:
Anyone wanna volunteer to implement this?

brookehus:
Done in #950

PR #950, opened 2016-11-30
add LandMarkRMSDFeaturizer

msultan:
 - [ ] Add tests
 - [ ] Update changelog

adds #935 . @mpharrigan could you pull and compare against your  implementation for correctness?


msultan:
One thing i am gonna note is that this is method of kernelizaing every possible metric is un-sustainable give the space of distance metrics and kernel options. I am not sure what a better option would be though.

msmbuilder/commands/featurizer.py
```diff
@@ -133,6 +133,27 @@ def _atom_indices_type(self, fn):
...
+                   "when loading reference trajectory".format(top))
+            assert os.path.exists(top), err
+        return md.load(fn, top=top)
+
+    def _atom_indices_type(self, fn):
+        if fn is None:
+            return None
+        return np.loadtxt(fn, dtype=int, ndmin=1)
+
+    def _sigma_type(self, val):
```
mpharrigan: this shouldn't be necessary

msmbuilder/featurizer/featurizer.py
```diff
@@ -313,6 +313,104 @@ def partial_transform(self, traj):
...
+    kernel.
+
+    Parameters
+    ----------
+    reference_traj : md.Trajectory
+        The reference conformations to superpose each frame with respect to
+    atom_indices : np.ndarray, shape=(n_atoms,), dtype=int
+        The indices of the atoms to superpose and compute the distances with.
+        If not specified, all atoms are used.
+    sigma: np.float , dtype=0.3nm
```
msultan: I can fix the text but i was more worried if the implementation was returning the right numbers.

mpharrigan:
Just my thoughts: not to be considered merge blockers

1) Related to your point above. Is a separate class what we should be doing here? We might need lots

1a) There's a lot of copy-pasted code from rmsdfeaturizer

1b) Maybe (i) use inheritance (ii) make `kernel` a new init arg on rmsdfeaturizer

2) describe features returning `N` copies of `atom_indices` seems silly

msultan:
>1b) Maybe (i) use inheritance

I think its done now 

>describe features returning N copies of atom_indices seems silly
 
yea but it is consistent with the other dataframes.

PR #957, opened 2017-01-19
Added DSSPFeaturizer

hannahw1:
 - [ ] Add tests
 - [ ] Update changelog


cxhernandez:
@msmbuilder/developers: thoughts on whether percentage H, E, and C is satisfactory or should it just output the raw DSSP results?

brookehus:
This looks cool! +1 for first PR. Some things to think about:

1. If you look at the files changed, the program you used made a bunch of changes outside the actual featurizer you implemented. Can you clean up the file to include just changes in the DSSPFeaturizer? You could create a fresh branch from master and then copy and paste your code into `featurizer.py`.

2. You should add tests to make sure this does what you want and accounts for edge cases- there are example featurizer tests in the relevant directory.

3. You will need to make this available on the command line and you can follow these instructions:
a) Go to msmbuilder/msmbuilder/commands
b) In featurizer.py, add DSSPFeaturizer here
c) At the bottom of featurizer.py, write a short class like the other featurizers have
d) In __init__.py (still in the commands subdirectory), add the newly created DSSPFeaturizerCommand to the .featurizer imports
Make sure when you do this your python interpreter is not changing other parts of the code.

4. @cxhernandez - Why not make it a user option whether to give the raw results or the percentages/counts?

5. Before merging we will need to changelog this, I'm flagging for v3.8.

6. If the featurizer is really new and different then you could consider adding an example notebook too!

msmbuilder/featurizer/featurizer.py
```diff
@@ -997,6 +1008,36 @@ def partial_transform(self, traj):
         return md.shrake_rupley(traj, mode=self.mode, **self.kwargs)
 
 
+class DSSPFeaturizer(Featurizer):
+    """Featurizer based on Dictionary of protein secondary structure
```
brookehus: The class docstring needs to be just one line for command line compatibility. You can then add line breaks and explain more. See the other featurizers for examples.

msmbuilder/featurizer/featurizer.py
```diff
@@ -374,15 +380,18 @@ def describe_features(self, traj):
...
         top = traj.topology
 
-        aind_tuples = [self.atom_indices for _ in range(self.sliced_reference_traj.n_frames)]
+        aind_tuples = [self.atom_indices for _ in range(
+            self.sliced_reference_traj.n_frames)]
         zippy = zippy_maker(aind_tuples, top)
 
-        zippy = itertools.product(["LandMarkFeaturizer"], ["RMSD"], [0.3], zippy)
+        zippy = itertools.product(["LandMarkFeaturizer"], [
+                                  "RMSD"], [0.3], zippy)
```
msultan: Actually could you change this 0.3 to self.sigma? I messed it up in my last PR.

cxhernandez: Make a new PR...

msmbuilder/featurizer/featurizer.py
```diff
@@ -997,6 +1008,39 @@ def partial_transform(self, traj):
...
+    Our implementation is based on DSSP-2.2.0, written by Maarten L. Hekkelman
+    and distributed under the Boost Software license.
+    """
+
+    def __init__(self):
+        pass
+
+    def partial_transform(self, traj):
+        dssp = md.compute_dssp(traj)
+        x = np.sum(dssp == 'H', axis=1) / float(dssp.shape[1])
```
mpharrigan: docstring says "H" "G" or "I", but the logic says only H

msmbuilder/featurizer/featurizer.py
```diff
@@ -997,6 +1008,39 @@ def partial_transform(self, traj):
...
+        y = np.sum(dssp == 'E', axis=1) / float(dssp.shape[1])
+        z = 1.0 - x - y
```
mpharrigan: Is it safe to do this? Maybe explicitly list things so people can tell if something is wrong (e.g. they don't add up to 1)

msmbuilder/featurizer/featurizer.py
```diff
@@ -280,11 +283,12 @@ def __init__(self, reference_traj=None, atom_indices=None, trj0=None):
...
         self.atom_indices = atom_indices
         if self.atom_indices is not None:
-            self.sliced_reference_traj = reference_traj.atom_slice(self.atom_indices)
+            self.sliced_reference_traj = reference_traj.atom_slice(
+                self.atom_indices)
         else:
             self.sliced_reference_traj = reference_traj
-            self.atom_indices = [i for i in range(self.sliced_reference_traj.n_atoms)]
-
+            self.atom_indices = [i for i in range(
```
mpharrigan: Can you get rid of these awkward line breaks

msmbuilder/featurizer/featurizer.py
cxhernandez: Default is simplified coding— so this should be fine.

PR #958, opened 2017-01-23
Ward clus centers

brookehus:
 - [x] Add tests
 - [x] Update changelog

Added `cluster_centers_` to `LandmarkAgglomerative` which computes the mean of all points in a cluster along every dimension

### DIFF
--- a/msmbuilder/cluster/agglomerative.py
+++ b/msmbuilder/cluster/agglomerative.py
@@ -142,8 +142,9 @@ class _LandmarkAgglomerative(ClusterMixin, TransformerMixin):
     ----------
     landmark_labels_
     landmarks_
-    cluster_centers_ : mean of all points in each cluster along every dimension,
-        unless metric == 'rmsd'
+    cluster_centers_ : array of cluster centers where the center of a cluster
+        is the mean of all points in that cluster along every dimension of
+        the original data. Returns None if metric == 'rmsd'.
     """
 
     def __init__(self, n_clusters, n_landmarks=None, linkage='average',
--- a/msmbuilder/cluster/agglomerative.py
+++ b/msmbuilder/cluster/agglomerative.py
@@ -159,7 +159,6 @@ class _LandmarkAgglomerative(ClusterMixin, TransformerMixin):
 
         self.landmark_labels_ = None
         self.landmarks_ = None
-
         self.cluster_centers_ = None
 
     def fit(self, X, y=None):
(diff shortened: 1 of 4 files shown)


brookehus:
I can get the double well test to work if I partition the double well into 4 bins along every dimension instead of 10 (see the last commit). I think 10 bins were getting zero populations/wonky transitions. Are @msmbuilder/developers okay with this change? Will assume so unless someone comments.

msmbuilder/cluster/agglomerative.py
```diff
@@ -142,6 +142,8 @@ class _LandmarkAgglomerative(ClusterMixin, TransformerMixin):
     ----------
     landmark_labels_
     landmarks_
+    cluster_centers_ : mean of all points in each cluster along every dimension,
```
cxhernandez: Does this mean you're using the centroid of all cluster centers?

msmbuilder/cluster/agglomerative.py
```diff
@@ -260,6 +263,13 @@ def predict(self, X):
             else:
                 print("No data points were assigned to cluster {}".format(i))
 
+        if self.metric != 'rmsd':
+            cluster_centers_ = []
+            for i in range(self.n_clusters):
+                temp = list(np.mean(self.landmarks_[self.landmark_labels_==i],axis=0))
```
cxhernandez: space before `axis`

msmbuilder/tests/test_agglomerative.py
```diff
@@ -113,3 +113,13 @@ def test_alanine_dipeptide():
     labels = model.fit_predict(trajectories[0][0:100])
 
     assert len(np.unique(np.concatenate(labels))) <= n_clusters
+
+
+def test_cluster_centers():
+    x = [random.randn(20, 2) + 10, random.randn(20, 2)]
+    n_clusters = np.random.randint(2,7)
```
cxhernandez: space after comma

msmbuilder/cluster/agglomerative.py
brookehus: I don't understand your question- can you rephrase? For each cluster, I am taking the average along each dimension and reporting the 'center' as the point defined by those averages.

cxhernandez: Ah okay, I misunderstood. Maybe just leave this as the default sklearn description for `cluster_centers_` (with a note about rmsd)? Something like:

>  Coordinates of cluster centers (unless RMSD is the metric)

Could you also follow the sklearn docstring template for attributes?

```
    Attributes
    ----------
    attribute_ : var_type, [dim1, dim2]
        Description here.
```

brookehus: done 😃

PR #960, opened 2017-01-24
Add reference to Husic et. al. in docs

brookehus:
- [ ] Implement feature / fix bug
 - [ ] Add tests
 - [ ] Update changelog

Add reference for Brooke's paper


mpharrigan:
I don't remember if you have to do anything to actually render changes to the bib file

mpharrigan:
Ah no we good https://github.com/msmbuilder/msmbuilder/blob/master/docs/conf.py#L305

PR #963, opened 2017-01-27
Post-release updates following v3.7.0

brookehus:
Travis build for conda-recipe passed so tentatively looks like .zip alone is ok


brookehus:
@mpharrigan can you look at my changes to the devtools readme

PR #975, opened 2017-03-24
Soft min contact feat

msultan:
 - [x] Add tests
 - [x] Update changelog

Added a contact featurizer that is capable of taking a soft minimization. This follows the commit on [mdtraj #1247](https://github.com/mdtraj/mdtraj/pull/1247). I  dont know if we can merge this until we cut a new mdtraj release and pin msmbuilder to the new version.

### DIFF
--- a/msmbuilder/featurizer/featurizer.py
+++ b/msmbuilder/featurizer/featurizer.py
@@ -16,6 +16,7 @@ import sklearn.pipeline
 from scipy.stats import vonmises as vm
 from msmbuilder import libdistance
 import itertools
+import inspect
 from sklearn.base import TransformerMixin
 from sklearn.externals.joblib import Parallel, delayed
 from ..base import BaseEstimator
@@ -1048,6 +1049,9 @@ class ContactFeaturizer(Featurizer):
         self.ignore_nonprotein = ignore_nonprotein
         self.soft_min = soft_min
         self.soft_min_beta = soft_min_beta
+        if self.soft_min and not 'soft_min' in inspect.signature(md.compute_contacts).parameters:
+            raise ValueError("Sorry but soft_min requires the latest version"
+                             "of mdtraj")
 
     def _transform(self, distances):
         return distances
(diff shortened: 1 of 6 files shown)


msultan:
also changed the contact featurizer dataframe to be less useless.

mpharrigan:
for compatibility with old mdtrajs, have something like
```python
if softmax:
  md.compute_contacts(x, y, softmax=True, beta=whatever)
else:
  md.compute_contacts(x,y)
```

so it will only break if you try to use the new feature (but don't have a new mdtraj)

msultan:
alright so i have now added exceptions to the tests until we get to the next release. The code in the transform method should then be modified at that point. This should be ready to merge after travis passes.

cxhernandez:
You're gonna need a try-except to throw an error about using softmin with an incompatible version

msultan:
@cxhernandez it already raises a value error if you try to setup the featurizer object without having the softmin option. I have disabled the tests for now though

msultan:
No idea why one of the travis builds keeps stalling. Appveyor is also broken for some other weird reason.

msultan:
Thanks a lot for looking into this @jchodera , gonna try pinning conda build to hopefully resolve the issue.

mpharrigan:
Can you split the test into two tests (copy and paste is good). One with old behavior (that is not skipped) and one with new behavior (skipped for now, you will be in charge of enabling upon new mdtraj)

msmbuilder/featurizer/featurizer.py
```diff
@@ -1029,12 +1030,28 @@ class ContactFeaturizer(Featurizer):
...
+    def __init__(self, contacts='all', scheme='closest-heavy', ignore_nonprotein=True,
+                 soft_min=False, soft_min_beta=20):
         self.contacts = contacts
         self.scheme = scheme
         self.ignore_nonprotein = ignore_nonprotein
+        self.soft_min = soft_min
+        self.soft_min_beta = soft_min_beta
+        if self.soft_min and not 'soft_min' in inspect.signature(md.compute_contacts).parameters:
+            raise ValueError("Sorry but soft_min requires the latest version"
+                             "of mdtraj")
```
cxhernandez: forgot a space here

Issue #979, opened 2017-03-29
HMM does not work in sklearn.pipeline.Pipeline


cxhernandez:
```python
pipeline = Pipeline([
                     ('decomp', tICA(n_components=2)),
                     ('hmm', HMM(n_states=6))
                    ])

assignments = pipeline.fit_transform(data)
```

results in:

```
TypeError                                 Traceback (most recent call last)
<ipython-input-10-41ecb70e3515> in <module>()
     12                     ])
     13 
---> 14 hmm_assignments = pipeline.fit_transform(data)

/home/server/miniconda3/envs/analysis/lib/python3.5/site-packages/sklearn/pipeline.py in fit_transform(self, X, y, **fit_params)
    305             return Xt
    306         else:
--> 307             return last_step.fit(Xt, y, **fit_params).transform(Xt)
    308 
    309     @if_delegate_has_method(delegate='_final_estimator')

TypeError: fit() takes exactly one argument (2 given)
```

cxhernandez:
done in #980

PR #980, opened 2017-03-29
pipeline compatibility for HMMs

cxhernandez:
 - [x] Add tests
 - [x] Update changelog

re: #979


cxhernandez:
Also realized that we need to make a `fit_predict` as well....

cxhernandez:
ping @mpharrigan: there was a random test failure

cxhernandez:
Also why does predict return a tuple with the score?

PR #984, opened 2017-04-06
Fix Badges

cxhernandez:
somehow spaces were added

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,10 +1,10 @@
 MSMBuilder
 ==========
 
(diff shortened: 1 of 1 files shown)


mpharrigan:
Good find (and very weird)

cxhernandez:
I found this issue with msmexplorer and osprey too. I wonder if GitHub has been tinkering....

mpharrigan:
That would be impossible, it would change the git hashes and break everything

mpharrigan:
although they changed their markdown rendering code recently :) https://githubengineering.com/a-formal-spec-for-github-markdown/

Issue #988, opened 2017-04-10
Add py36 to CI tests


cxhernandez:
@msmbuilder/developers: How do y'all feel about adding another test to the matrix? Should we drop testing on py27 as well?

brookehus:
I think 3.6 is good but suspect there are still many users with 2.7, wouldn't drop those tests until we formally end support for 2.7

cxhernandez:
also related to https://github.com/msmbuilder/msmbuilder/issues/852

cxhernandez:
Unfortunately, I can't find anything related to when `sklearn` might drop py2 support. But many important packages, like `pandas` and `matplotlib`, have signed the petition @mpharrigan mentioned.

cxhernandez:
done in #990

Issue #852, opened 2016-07-08
Pledge to drop support for python 2


mpharrigan:
http://python3statement.github.io/

We should sign this pledge!

cxhernandez:
+1

mpharrigan:
I'd actually like to do this much sooner. Maybe the release-after-next (i.e. 3.7). comments?

rmcgibbo:
-1 on any pledge.

jchodera:
Why drop py27 support sooner?

msultan:
I propose v4.0+ given that we are doing version releases every quarter. That should be 15ish months from now. That should hopefully allow for a smoother transition.

cxhernandez:
I think we cater to a small enough community that signing this pledge is okay.... Also, the sooner we sign the more publicity we get (who knows how many names will be on that list in a year)

mpharrigan:
> Why drop py27 support sooner?

Like the site says, supporting multiple versions of python "is a small but constant friction in the development of a lot of code". I haven't used python 2 other than to test for compatibility issues in the past two years. I don't know about the other @msmbuilder/developers but I think a lot of them use py3 primarily as well. I don't want subtle bugs to sneak in under py2 that we don't catch because no one uses it, while also promising users that we support py2.

A bigger reason why I want to drop py2 is to give back to the python community. The fractionization between 2 and 3 is not good for anyone. I fully blame the python developers for making choices that guaranteed this transition would be terrible, but I also think it would be foolish to go back to python 2 and eventually people will move to python 3. Anything I can do to speed that process up, I'm all for.

Minor reason 1 is: compatibility checks and `six` reduce the readability of the code. Python's greatest strength (imo) is readability. MSMBuilder code is read more than written

Minor reasons 2: there are some features that I've wished I could use. Automatic new-style classes, `TemporaryDirectory` as a context manager, `os.makedirs(exist_ok=True)`, subprocess improvements (`run` and `subprocess.DEVNULL`) just off the top of my head. There's also the matrix multiply operator, type hinting, and coroutines (although I don't think we're in any hurry to use those particular things)

kyleabeauchamp:
There's a lot of legacy python applications that are not going to be migrated to py3 on the year timescale.  I think we should ask ourselves "will a sizable fraction of our users be negatively impacted by dropping py27"?  I'd only be willing to sign a pledge once the answer to that question is no.  

[...]

PR #989, opened 2017-04-11
Update .travis.yml

cxhernandez:
add py3.6 in testing #988

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -24,6 +24,7 @@ env:
     - CONDA_PY=2.7 CONDA_NPY=1.11
     - CONDA_PY=3.4 CONDA_NPY=1.10
     - CONDA_PY=3.5 CONDA_NPY=1.11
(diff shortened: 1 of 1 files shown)


mpharrigan:
Should we remove python 3.4? Should we add numpy 1.12?

cxhernandez:
> Should we remove python 3.4?

Why 3.4 before 2.7?

> Should we add numpy 1.12?

👍

mpharrigan:
The general idea is "latest two pythons plus python 2.7". I'm sure usage statistics would back this up as would amount of discussion #852

cxhernandez:
Ah, got it. Should we look into implementing #852?

mpharrigan:
Feel free to continue the discussion there.

Another point worth mentioning: if the code passes tests in python 3.5, it's likely it will also work in 3.4 whereas 2.7 is the last of its kind, and removing it isn't just a normal matter of business

PR #990, opened 2017-04-11
Py36 testing

cxhernandez:
moving #989 to here

re: #988

### DIFF
--- a/docs/requirements.txt
+++ b/docs/requirements.txt
@@ -1,6 +1,7 @@
 numpydoc
 matplotlib
-ipython-notebook
+jupyter
+notebook
 jinja2
 openmm
 nbconvert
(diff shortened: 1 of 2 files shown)


cxhernandez:
@mpharrigan: should we build the docs in 3.6 now?

mpharrigan:
> @mpharrigan: should we build the docs in 3.6 now?

sure

Why did you make a new PR?

Also, we could keep the py3.5 on numpy 1.11 to test against multiple versions of numpy. Our three configurations would be loosely: a python2 user, a slow-to-upgrade python3 user, a bleeding-edge user

cxhernandez:
> Why did you make a new PR?

GitHub didn't let me push to this repo remotely

docs/requirements.txt
```diff
@@ -1,6 +1,7 @@
 numpydoc
 matplotlib
-ipython-notebook
+jupyter
+notebook
```
mpharrigan: is this right? notebook?

cxhernandez: https://pypi.python.org/pypi/notebook

Issue #994, opened 2017-04-13
Warning/error in tICA when N>p


rmcgibbo:
With the default shrinkage covariance matrix estimator, no warning or error is thrown when the number of data points, N, is less than the number of features. But very frequently, the largest eigenvalue can be greater than 1. Without the shrinkage estimator, you'd get a linear algebra error from the covariance matrix not being positive definite in this situation. Perhaps a warning or error would be appropriate here.

```
In [177]: tICA(n_components=1).fit([np.random.RandomState(0).randn(9,10)]).eigenvalues_                                                                                    
Out[177]: array([ 1.26276845])

In [178]: tICA(n_components=1).fit([np.random.RandomState(0).randn(9,10)]).timescales_                                                                                     
Out[178]: array([-4.28620728])
```

cxhernandez:
done in #998

PR #998, opened 2017-04-24
tica warning

cxhernandez:
re: https://github.com/msmbuilder/msmbuilder/issues/994


msultan:
wouldn't sparse tica or pca or other decomposition methods have similar issues? Maybe just add it to the super class?

cxhernandez:
This is the superclass for `SparseTICA`....

PR #1005, opened 2017-04-29
Added sampling error attribute to MSM object

jadeshi:
 - [ ] Add tests
 - [ ] Update changelog

MSM object now has an attribute called transmat_error_, which is a matrix of standard errors computed for each transition probability using the counts matrix.


cxhernandez:
?

PR #1008, opened 2017-04-29
test to run before running all nosetests

brookehus:
- [ ] Implement feature / fix bug
 - [x] Add tests
 - [ ] Update changelog

I am not sure if this is the best solution (or if anything is needed at all), but optionally a user could run this test before running any other test to make sure that everything's in place.

Will this mess up windows users by requiring `cvxpy` (based on the comment [here](https://github.com/msmbuilder/msmbuilder/blob/e6839bccd3f1c49f0d1132375a95de61fd0914be/devtools/conda-recipe/meta.yaml#L47))?


msmbuilder/tests/test_dependencies.py
```diff
@@ -0,0 +1,27 @@
...
+    test_dependencies = ['munkres', 'numdifftools', 'statsmodels', 'hmmlearn', 'cvxpy']
+
+    for td in test_dependencies:
+        if td not in package_names:
+            raise RuntimeError('Please install {} to continue'.format(td))
+
+def test_msmbuilder_data():
+    home = os.path.expanduser('~')
+    try:
+        os.stat(home+'/msmbuilder_data/')
```
cxhernandez: Use [`os.path.join`](https://docs.python.org/3/library/os.path.html#os.path.join) instead of adding strings.

brookehus:
This seems to be not immediately friendly with the builds so maybe just better to assume people can look at the devtools readme or conda recipe to figure it out. It would be nice if failures such as `No such file: /Users/bhusic/msmbuilder_data/alanine_dipeptide/ala2.pdb` were more informative (i.e. instruct users to install `msmb_data`), but not sure where the best place to insert that is. Any thoughts?

cxhernandez:
You you do something like:

```
try:
    from msmbuilder.example_datasets import FsPeptide
    fspeptide = FsPeptide().get()
except:
    warnings.warn('Dataset not found. Did you install msmb_data?')
```

mpharrigan:
I don't think those tests will ever fail. `get` will download if necessary. `get_cached` will fail without the msmb_data package. The tests use the latter

cxhernandez:
Ah, good point.

brookehus:
I see at least 3 options:

1. Test `get_cached` instead of `get`
2. Change tests that use `get_cached` to use `get` so this problem will never come up
3. Do nothing

Thoughts?

cxhernandez:
4. Test using [`has_msmb_data`](https://github.com/msmbuilder/msmbuilder/blob/master/msmbuilder/example_datasets/base.py#L208-L223) (I forgot about this)

PR #1015, opened 2017-05-16
Correct tica warning

dr-nate:
Fix confusing ~error~ warning message. Currently it gives a message like:

```
The number of features (500) is greater than the length of the data (2172)...
```
When the number of features is 2172 and the length of the data is 500.

 - [ ] Add tests
 - [ ] Update changelog

[Describe changes here]


dr-nate:
Note that the actual test in the if statement is correct, just the warning is printed incorrectly. Minor but why not fix it.

mpharrigan:
lol thanks nate!

cxhernandez:
woops 😞

PR #1018, opened 2017-06-11
Change to FeatureSelector 'which_feat' setter. 

RobertArbon:
 - [x] Add tests
 - [x] Update changelog

`FeatureSelector` setter for `which_feat` previously cast all non-list variables to a list. 
```
if not list:
    value = [value]
```
This was causing `KeyError`'s with the  Tree-based Parzen Estimator method in Osprey as it passes tuples (seemingly impossible to change) to the `which_feat` variable. e.g.
```
which_feat = ('phi', 'psi')
```
got converted by the setter to:
```
which_feat = [('phi', 'psi')]
```
instead of:
```
which_feat = ['phi', 'psi']
```

This PR now fixes that by checking for string variable separately and casts everything else as a list: 
```
if string:
    value = [value]
if dictionary:
   TypeError
if value not in feature_list:
    TypeError
else:
     value = list(value)
```

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -17,6 +17,7 @@ New Features
 
 Improvements
 ~~~~~~~~~~~~
+- ```FeatureSelector``` is now compatible with Tree-structure Parzen Estimator method in Osprey
 
 
 v3.8 (April 26, 2017)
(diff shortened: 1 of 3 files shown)


cxhernandez:
Thanks!

PR #1019, opened 2017-06-15
Fixed error in ligand contact CA featurizer

cing:
 - [x] Add tests

Just a tiny typo that prevented ligand contact featurizer scheme="ca" from running. I made a test that catches it but it required a new fake trajectory with only one c-alpha per residue.

Bigger picture, the single-atom 'ca' scheme of ligand contact featurizing should probably support any atom name in the ligand rather than strictly 'ca'. Until then, renaming an atom in the ligand topology is pretty painless.


brookehus:
Looks good, thanks. If we were going the change anything about the scheme parameter, it should be done in mdtraj (see [here](https://github.com/mdtraj/mdtraj/blob/master/mdtraj/geometry/contact.py#L157) and following).

PR #1027, opened 2017-08-03
Add a maximum number of iterations in prinz mle

mpharrigan:
For some reason, my mle estimation was not converging within tolerance which resulted in a seemingly infinite loop and you cant ctrl-C. This adds a limit to the number of iterations attempted. 

looks like this pr is also pulling in #1024 . i can rebase if necessary

### DIFF
--- a/msmbuilder/msm/src/transmat_mle_prinz.c
+++ b/msmbuilder/msm/src/transmat_mle_prinz.c
@@ -4,6 +4,8 @@
 #include <stdlib.h>
 #include "transmat_mle_prinz.h"
 
+#define MAX_ITER 10000
+
 /**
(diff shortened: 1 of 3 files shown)


msmbuilder/msm/src/transmat_mle_prinz.c
```diff
@@ -4,6 +4,8 @@
 #include <stdlib.h>
 #include "transmat_mle_prinz.h"
 
+#define MAX_ITER 10000
```
msultan: Is that enough iterations to fit most matrices? I don't have an intuition about what the order of magnitude for the iteration count should be.

mpharrigan: it's usually 5 - 50

PR #1028, opened 2017-08-07
I was getting a "chain_dict referenced before assignment" error here.

dr-nate:
It's unclear to me if there was ever any reason for `chain_dict` to not
be always initialized at the start of this method. It doesn't seem to
be used anywhere else.

 - [ ] Add tests
 - [ ] Update changelog

[Describe changes here]


brookehus:
Thanks for this! Changes make sense to me (oops), will wait for builds to pass.

brookehus:
@msmbuilder/developers can i merge? what's up with travis?

cxhernandez:
yeah go ahead

PR #1032, opened 2017-08-23
Update gmrq.rst

cxhernandez:
Add a warning about using GMRQ for lag time selection.


cxhernandez:
We should also link to @mpharrigan's blog post: http://mpharrigan.com/2016/09/29/gmrq.html

brookehus:
should we add that you also can't compare across number of timescales scored or is that so obvious it does not require comment

cxhernandez:
Probably should include it just in case...

PR #1034, opened 2017-08-23
Fix Conda Build errors

cxhernandez:
This fixed things locally for me. Hopefully, this isn't compromising the build workflow in any way.

### DIFF
--- a/devtools/conda-recipe/meta.yaml
+++ b/devtools/conda-recipe/meta.yaml
@@ -18,7 +18,7 @@ requirements:
     - setuptools
     - cython
     - numpy x.x
-    - mdtraj
+    - mdtraj <=1.8
   run:
     - python
     - setuptools
@@ -26,7 +26,7 @@ requirements:
     - scipy
     - pandas
     - six
-    - mdtraj
+    - mdtraj <=1.8
     - scikit-learn
     - numpydoc
     - pytables
--- a/devtools/conda-recipe/meta.yaml
+++ b/devtools/conda-recipe/meta.yaml
@@ -24,7 +24,7 @@ requirements:
     - setuptools
     - numpy x.x
     - scipy
-    - pandas
+    - pandas <0.20
     - six
     - mdtraj <=1.8
     - scikit-learn
--- a/devtools/travis-ci/build_docs.sh
+++ b/devtools/travis-ci/build_docs.sh
@@ -12,7 +12,7 @@ conda install -yq --use-local msmbuilder
 # sphinx_rtd_theme's latest releases are not available
 # neither is msmb_theme
 # neither is sphinx > 1.3.1 (fix #1892 autodoc problem)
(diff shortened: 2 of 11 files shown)


cxhernandez:
yayy they finally run!

cxhernandez:
There's a lot of errors from the GHMM code. Was this part of the code changed recently?

mpharrigan:
why did you remove numpy x.x ? that's there for a reason

cxhernandez:
Locally, that was giving me a deprecation warning with the latest version of conda-build.

mpharrigan:
Do you want to investigate what the preferred solution is to depend on a specific numpy ABI if not `numpy x.x`?

cxhernandez:
https://conda.io/docs/building/variants.html#customizing-compatibility

cxhernandez:
https://conda.io/docs/user-guide/tasks/build-packages/variants.html#conda-variables-and-command-line-arguments-to-conda-build

>To ensure consistency with existing users of conda-build, environment variables such as CONDA_PY behave as they always have, and they overwrite all variants set in files or passed to the API.

Does this mean we might not need this if those variables are set?

cxhernandez:
Yeah, just checked locally. This seems to work.

cxhernandez:
UPDATE: I don't care enough to understand the new system. I'm pinning to conda-build v2 (which conda-forge uses) and hoping for the best.

cxhernandez:
Okay seems like there's some issues with HMMLearn. `model.sample(N)` results in an array of `None` objects.

cxhernandez:
The only reason `hmmlearn` v0.1.1 was building on omnia was because it didn't have any unit tests.... the latest version fails for 4 and throws errors in 39 using `numpy` 1.12. I vote we get rid of this as a test dependency, if possible.

msultan:
Can we block the ghmm test for the time being and merge this? There are several PRs that are being blocked by this.

mpharrigan:
also I realized we don't need `numpy x.x` to run the tests because it will automatically test against the same numpy version that was used to build (if you keep the environment variable)

AND conda-forge has a new cockamamy scheme that doesn't pin to numpy and use `x.x`

brookehus:
+1 to @msultan's comment

brookehus:
@msmbuilder/developers, when @cxhernandez gets back let's prioritize getting this fixed

cxhernandez:
Now there are issues with MDTraj testing data having been removed. Working on this now...

cxhernandez:
Looks like someone broke RMSD in mdtraj v1.9. When I pin to 1.8.0, the tests work.

cxhernandez:
yay!

cxhernandez:
It seems that `Win` doesn't have the latest version of `statsmodels` yet (re: https://github.com/statsmodels/statsmodels/issues/3752).

msultan:
Thank you @cxhernandez !

cxhernandez:
So the GHMM tests just required `startprob_` to be initialized. Apparently between versions, this became necessary for sampling from the GHMM class -__-

PR #1035, opened 2017-08-28
KL, JS divergence utilities compatible with agglomerative clustering

brookehus:
 - [x] Add tests
 - [x] Update changelog

Multiple implementations of Kullbeck-Leibler, symmetric KL, and Jenson-Shannon divergences for comparing probability distributions. A manual method is used to account for zero entries which are common in MSM analyses without pseudocounts, but the scipy method can be selected. Array methods for computing the divergence between two elements of an array, and "MSM" methods for computing the divergence between two MSMs (which must be flattened before analysis so they don't look like lists of arrays) are also available, and are designed to be used as custom distance functions with `LandmarkAgglomerative` clustering. There are lots of tests.

I also cleaned up the changelog by referencing PRs where they hadn't been referenced.


brookehus:
I suppose I could rebase this

msultan:
Looks good to me though i would have thought this could have gone into ```mdentropy```?

cxhernandez:
Go ahead and merge, @brookehus (I can't override the tests)

Issue #655, opened 2015-09-23
BACE in msmbuilder 3.0


juanmit:
I am trying to use BACE method to lump the micro states into macro states. And I realize that BACE has not been built in msmbuilder 3.0. So I used BACE method in msmbilder 2.8 and get a set of .dat file. I am wondering how can I use these .dat files to build a Markov state model in msmbuilder 3.0? In other words. I am wondering that are there any ways I can convert these .dat files to a sequence for msmbuilder 3.0. Thanks a lot for helping.

mpharrigan:
Is there a reason you don't want to use pcca+? 

Note for devs: maybe this is implemented in msmtools?

juanmit:
Thanks for replying. 
Because one of the tutorials online is saying that PCCA+ will lead to poor macrostates so I am trying to use BACE instead of PCCA+. Is PCCA+ working well on msmbuilder 3 now?

jchodera:
Wouldn't it be useful to have all of these methods available for comparison purposes?

Devs: What about working with the @franknoe folks to pull a full suite of lumping tools into [msmtools](https://github.com/markovmodel/msmtools) and utilizing that in both [`msmbuilder`](https://github.com/msmbuilder/msmbuilder) and [`pyemma`](https://github.com/markovmodel/PyEMMA)?

mpharrigan:
Looks like msmtools has pcca+. I couldn't find bace

Yes, @jchodera that is how we would go about it

mpharrigan:
@juanmit - to my knowledge, pcca+ has been the "standard" method for some time (including in msmb 2.8 tutorial)

juanmit:
Thanks for helping. I  think I will use pcca+ for my current model.

cxhernandez:
bace is implemented in #1050

PR #1050, opened 2017-11-06
Implementation of BACE

brookehus:
 - [x] Add tests
 - [x] Update changelog

This moves BACE over to MSMB3 (finally!). I prioritized fidelity to MSMB2 results. I removed the option for sparse matrices that MSMB2 had. The distance matrix has its reciprocal taken twice for the old sparse matrix formulation, and I didn't remove this, so there will be some precision issues. However, these precision issues were present in the original version of BACE so it will return the same result.

The code is as much copied and pasted from Greg's code as possible.

### DIFF
--- a/msmbuilder/lumping/bace.py
+++ b/msmbuilder/lumping/bace.py
@@ -1,0 +1,340 @@
+# Author: Brooke Husic <brookehusic@gmail.com>
+# Contributors: Greg Bowman
+# Copyright (c) 2017, Stanford University and the Authors
+# All rights reserved.
+
+from __future__ import print_function, division, absolute_import
+
+import numpy as np
+from msmbuilder.msm import MarkovStateModel
+import functools
+import multiprocessing
+import os
+
+class BACE(MarkovStateModel):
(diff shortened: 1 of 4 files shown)


brookehus:
cc @nhstanley

nhstanley:
Nice! Will allow us to close #655

msmbuilder/lumping/bace.py
```diff
@@ -0,0 +1,380 @@
...
+                i = 0
+                while dest.shape[0] > i:
+                    if i + self.chunk_size > dest.shape[0]:
+                        indices.append((s, dest[i:]))
+                    else:
+                        indices.append((s, dest[i:i + self.chunk_size]))
+                    i += self.chunk_size
+        return indices
+
+    def mergeTwoClosestStates(self, c, w, indRecalc, dMat,
```
cxhernandez: Should class functions such as this one be hidden from users (e.g. `_mergeTwoClosestStates`)?

brookehus: yeah that's a good idea I'll hide them all

cxhernandez: 👍 LGTM otherwise

PR #1062, opened 2018-02-12
SparseTICA supports commute mapping

jeiros:
Added a commute_mapping attribute to the SparseTICA class so that the `transform` method of the parent tICA class is used.

 - [x] Add tests
 - [x] Update changelog

Since `tICA` can now weight the eigenvectors with commute mapping, thought it made sense to add it to the `SparseTICA` class too.

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -37,6 +37,7 @@ offset CA atoms.
   
 Improvements
 ~~~~~~~~~~~~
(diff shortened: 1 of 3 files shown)


cxhernandez:
Thanks @jeiros! will merge once https://github.com/msmbuilder/msmbuilder/pull/1063 finishes building.

PR #1091, opened 2018-06-02
Fix vmfeat

cxhernandez:
fixed an issue where the vm featurizer's `loc` parameter wouldn't set properly. mostly an issue for osprey users


cxhernandez:
Something happened in cython 0.28 which breaks msmbuilder, will need to investigate

cxhernandez:
looks like pip 10 has a similar issue

